PR #75, opened 2026-06-02
Support controller as string or array of strings (#47); v3.0.0 + dependency updates

### LINKED ISSUE #47: Support controller as string or array of strings per ZCAP-LD spec
## Summary

The live W3C CCG ZCAP-LD v0.3 draft allows `controller` to be either a single string or an array of strings on both root and delegated zcaps. The current model only supports a single string, and verifier authorization only checks that one value. This means valid spec-conformant zcaps with multiple controllers cannot round-trip or verify correctly.

## Spec Requirement

Root zcap section:

> A root zcap MUST have a `controller` that is a string or an array of strings that each express a URI that identifies a controller for the root zcap.

Delegated zcap section:

> A delegated zcap MUST have a `controller` that is a string or an array of strings that each express a URI that identifies a controller for the delegated zcap.

In the fetched spec snapshot used during review, these appeared around lines 145 and 213 respectively.

## Current Behavior

`Capability.Controller` is currently modeled as a single string:

- `src/ZcapLd.Core/Models/Capability.cs`

```csharp
[JsonPropertyName("controller")]
public string Controller { get; set; } = string.Empty;
```

Authorization also assumes a single controller string:

- `src/ZcapLd.Core/Services/VerificationService.cs`

```csharp
private bool IsControllerAuthorized(string verificationMethod, Capability capability)
{
    var did = verificationMethod.Split('#')[0];
    return did == capability.Controller || verificationMethod == capability.Controller;
}
```

Related service paths also pass/compare single controller values:

- `CapabilityService.CreateRootCapabilityAsync(string controller, ...)`
- `CapabilityService.DelegateCapabilityAsync(..., string newController, ...)`
- signing uses `parentCapability.Controller` as the signer DID for delegation

## Why This Matters

A valid root zcap such as this is allowed by the draft:

```json
{
  "id": "urn:zcap:root:https%3A%2F%2Fexample.com%2Fresource",
  "controller": [
    "did:key:z6MkAlice",
    "did:key:z6MkBob"
  ],
  "invocationTarget": "https://example.com/resource"
}
```

With the current model, this shape cannot deserialize into `Capability` because `controller` is not a string. Even if a consumer manually adapts the model, authorization currently only checks one controller value.

Similarly, valid delegated zcaps with multiple controllers should permit invocation/delegation by any authorized controller in the array, subject to proof purpose and verification method rules.

## Expected Behavior

- Root and delegated zcaps should accept `controller` as either:
  - a single URI string, or
  - an array of URI strings.
- Serialization should preserve the input shape where possible, or use a documented canonical representation that remains accepted by cross-language verifiers.
- Authorization should succeed when the proof verification method matches any controller in the effective controller set.
- Validation should reject malformed controller values:
  - missing / empty string
  - empty array
  - array containing non-strings or empty strings
  - values that are not valid URI-like controller identifiers, as far as the library can reasonably validate.

## Suggested Implementation Approach

Introduce a small value type/converter rather than sprinkling `object` checks throughout the codebase. For example:

```csharp
[JsonConverter(typeof(ControllerSetJsonConverter))]
public sealed class ControllerSet
{
    public IReadOnlyList<string> Values { get; }
    public bool ContainsVerificationMethod(string verificationMethod) { ... }
}
```

Then update the model:

```csharp
[JsonPropertyName("controller")]
public ControllerSet Controller { get; set; } = ControllerSet.Empty;
```

Alternative: keep a compatibility-facing `string Controller` for simple creators and add a serialized backing property, but that risks preserving the same single-controller assumptions. A dedicated type would make the spec shape explicit.

Areas likely affected:

- `src/ZcapLd.Core/Models/Capability.cs`
- `src/ZcapLd.Core/Services/VerificationService.cs`
- `src/ZcapLd.Core/Cryptography/ProofSigningPayloadBuilder.cs` if canonicalization needs a stable serialized shape
- tests in `tests/ZcapLd.Core.Tests/Compliance/` and `tests/ZcapLd.Core.Tests/Models/`

## Tests to Add

1. Model serialization/deserialization:
   - deserialize root zcap with `controller` as a string
   - deserialize root zcap with `controller` as an array
   - serialize single-controller capability
   - serialize multi-controller capability
   - reject empty controller array

2. Root invocation authorization:
   - root zcap controller array includes Alice and Bob
   - invocation signed by Alice verifies
   - invocation signed by Bob verifies
   - invocation signed by Mallory fails

3. Delegation authorization:
   - parent capability has controller array `[Alice, Bob]`
   - child delegated zcap signed by Alice verifies
   - child delegated zcap signed by Bob verifies
   - child delegated zcap signed by Mallory fails

4. Cross-language canonicalization regression:
   - pin JCS bytes for a capability whose `controller` is an array
   - verify sign-time bytes match wire-time bytes

## Acceptance Criteria

- `controller` string and string-array forms are both accepted for root and delegated zcaps.
- Verification method authorization checks all controllers in the array.
- Existing single-controller APIs remain ergonomic or have a clear migration path.
- Compliance tests cover root and delegated controller arrays.
- Full test suite passes.

## Related Compliance Review Note

This was found during a source-of-truth re-check against the live W3C CCG draft. The repo's current compliance tests pass, but they do not yet cover the valid multi-controller shape.

moisesja:
Closes #47.

## Summary
Implements W3C ZCAP-LD v0.3 multi-controller support — `controller` may be a single URI string **or** an array of URI strings on root and delegated zcaps — and ships it as a **3.0.0** release together with dependency updates.

## Issue #47 — `controller` string | array
- New **`ControllerSet`** immutable value type (`Values`, `Primary`, `Count`, `IsEmpty`, `IsArrayForm`, `Contains`, `ContainsVerificationMethod`) with implicit conversions from `string` and `string[]`, so `Controller = "did:..."` and `Controller = new[]{...}` both compile.
- New internal **`ControllerSetJsonConverter`** that **preserves the on-wire shape**: a single controller stays a bare JSON string; an array stays an array (even single-element). This keeps JCS canonical bytes byte-stable for cross-language verifiers and means existing single-controller signatures/JCS pins are unchanged. Rejects malformed values (empty array, non-string entry, empty/whitespace string).
- `Capability.Controller` type changes `string` → `ControllerSet` (**breaking API**; assignments from `string`/`string[]` still compile, reads use `.Primary` / `.Values`).
- `CapabilityService` / `ICapabilityService` create + delegate take `ControllerSet`; `DelegateCapabilityAsync` gains an optional `signerDid` so any one of a multi-controller parent's controllers can sign (validated against the set).
- `VerificationService` authorization now matches **any** controller via `ContainsVerificationMethod`; empty-controller guards updated.

## Version → 3.0.0
- `ZcapLdVersion` `2.1.1` → `3.0.0` (SemVer-correct for the breaking `Controller` API change). CHANGELOG updated.

## Dependency updates
| Package | From | To |
|---|---|---|
| NetDid.Core / NetDid.Method.Key | 1.1.2 | 1.3.0 |
| Microsoft.SourceLink.GitHub | 10.0.201 | 10.0.300 |
| Microsoft.NET.Test.Sdk | 18.4.0 | 18.6.0 |
| FluentAssertions | 8.9.0 | 8.10.0 |
| coverlet.collector | 6.0.0 | 10.0.1 |
| Microsoft.Data.Sqlite | 10.0.6 | 10.0.8 |

### P-256 IEEE P1363 fix (NetDid 1.3.0)
NetDid 1.3.0 made **DER the default** for its legacy ECDSA `Sign`/`Verify` overloads (was P1363 in ≤1.2.0). The W3C `ecdsa-2019` / `EcdsaSecp256r1Signature2019` wire format requires **IEEE P1363** (fixed-width `r‖s`, 64 bytes for P-256), so emitting DER would silently break P-256 verification against peer Data Integrity stacks. `CryptoSuite` now calls NetDid's format-aware overloads with `EcdsaSignatureFormat.IeeeP1363`. Ed25519 (the default suite) is unaffected. Guarded by `CryptoSuiteTests.P256_Sign_ProducesIeeeP1363WireFormat_NotDer`.

## Tests
- New: `ControllerSetTests`, `MultiControllerAuthorizationTests` (Alice/Bob verify ✅, Mallory rejected ❌ — for both invocation and delegation), `CrossLanguageJcsInteropTests` array-shape pins, P-256 P1363 regression test.
- Migrated existing `.Controller` assertions to `.Primary` / `.IsEmpty`.
- **Full suite: 334 passed / 0 failed.**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,22 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/),
 and this project adheres to [Semantic Versioning](https://semver.org/).
 
-## [2.1.1] - Unreleased
+## [3.0.0] - Unreleased
+
+### Added
+
+- `controller` may now be a single URI string **or** an array of URI strings on both root and delegated zcaps, per W3C ZCAP-LD v0.3 (Issue #47). Authorization succeeds when a proof's verification method matches **any** controller in the set.
+- `ControllerSet` (in `ZcapLd.Core.Models`) — immutable value type modeling one or many controllers. Exposes `Values`, `Count`, `IsEmpty`, `IsArrayForm`, `Primary`, `Contains`, and `ContainsVerificationMethod`. Implicit conversions from `string` and `string[]` keep single-controller call sites ergonomic (`Controller = "did:..."` still compiles). Carries a `[JsonConverter]` that **preserves the on-wire shape** (a single controller stays a bare string; an array stays an array, even a single-element one) so JCS canonical bytes round-trip byte-stably for cross-language verifiers.
+- `ControllerSetJsonConverter` (internal) — reads/writes both wire shapes and rejects malformed controller values (empty array, non-string entry, empty/whitespace string) with `JsonException`.
+- `CapabilityService.DelegateCapabilityAsync` gains an optional `signerDid` parameter so any one of a multi-controller parent's controllers can sign the delegation (defaults to the parent's first controller; validated against the parent's controller set).
+
+### Changed
+
+- **BREAKING (API)**: `Capability.Controller` changes type from `string` to `ControllerSet`. Reads that previously treated it as a string need `.Primary` (the single/first controller) or `.Values`; assignments from a `string` or `string[]` continue to compile via implicit conversion. `ICapabilityService.CreateRootCapabilityAsync` / `DelegateCapabilityAsync` now take `ControllerSet` (string callers unaffected).
+- Single-controller capabilities are unchanged on the wire — `controller` still serializes as a bare string, so existing signatures and pinned JCS bytes are preserved.
+- New `CrossLanguageJcsInteropTests` pins lock the multi-controller shape: an array-form `controller` produces sign-time bytes byte-equal to what a peer verifier JCS-canonicalizes over the array-shaped wire body, and a single controller still emits a bare string (no one-element-array collapse).
+- Dependency updates: `NetDid.Core` / `NetDid.Method.Key` 1.1.2 → 1.3.0, `Microsoft.SourceLink.GitHub` 10.0.201 → 10.0.300. Test/example tooling: `Microsoft.NET.Test.Sdk` 18.4.0 → 18.6.0, `FluentAssertions` 8.9.0 → 8.10.0, `coverlet.collector` 6.0.0 → 10.0.1, `Microsoft.Data.Sqlite` 10.0.6 → 10.0.8.
+- P-256 signing/verification now explicitly requests IEEE P1363 from NetDid's format-aware overloads (`EcdsaSignatureFormat.IeeeP1363`). NetDid 1.3.0 changed the *default* of its legacy ECDSA `Sign`/`Verify` overloads from P1363 to ASN.1 DER; the W3C `ecdsa-2019` / `EcdsaSecp256r1Signature2019` wire format requires P1363 (fixed-width `r‖s`, 64 bytes for P-256), so `CryptoSuite` pins the format to keep `proofValue` interoperable with peer Data Integrity stacks. Ed25519 (the default suite) is unaffected. Regression-guarded by `CryptoSuiteTests.P256_Sign_ProducesIeeeP1363WireFormat_NotDer`.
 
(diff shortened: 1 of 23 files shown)


moisesja (commented):
There is no example on how to use multiple controllers

PR #77, opened 2026-06-02
fix: reject forbidden extra fields on root zcaps during validation (#49)

### LINKED ISSUE #49: Reject forbidden extra fields on root zcaps during validation
## Summary

The live W3C CCG ZCAP-LD v0.3 draft says a root zcap must contain only the root fields: `@context`, `id`, `controller`, and `invocationTarget`. The current creation path emits that shape correctly, but `CapabilityService.ValidateCapabilityAsync` is too permissive and can accept root capabilities that contain forbidden extra fields such as `allowedAction`, `caveat`, or unknown extension fields.

## Spec Requirement

In the **Root Capability** section, the draft defines the root shape and then states:

> A root zcap MUST NOT have any other fields.

In the fetched spec snapshot used during review, this appears around line 147.

The permitted root fields are:

- `@context`
- `id`
- `controller`
- `invocationTarget`

Root capabilities do not have a delegation proof, expiration, parent capability, caveats, allowed actions, or extension fields.

## Current Behavior

`CreateRootCapabilityAsync` emits the correct strict root shape:

```csharp
var capability = new Capability
{
    Context = "https://w3id.org/zcap/v1",
    Id = rootId,
    Controller = controller,
    InvocationTarget = invocationTarget,
    // AllowedAction, Caveat, Expires, ParentCapability, Proof intentionally unset.
};
```

However, `ValidateCapabilityAsync` only rejects three forbidden root fields:

```csharp
if (capability.Proof != null || capability.Expires != null || capability.ParentCapability != null)
{
    return Task.FromResult(false);
}
```

It does **not** reject:

- `AllowedAction`
- `Caveat`
- `AdditionalProperties` / unknown extension fields

The model explicitly preserves unknown fields via:

```csharp
[JsonExtensionData]
public Dictionary<string, JsonElement>? AdditionalProperties { get; set; }
```

That is important for delegated/canonicalization interop, but root validation still needs to reject those fields because root zcaps have an exact field set.

### Root with `allowedAction`

```json
{
  "id": "urn:zcap:root:https%3A%2F%2Fexample.com%2Ffoo",
  "controller": "did:key:example",
  "invocationTarget": "https://example.com/foo",
  "allowedAction": ["read"]
}
```

### Root with `caveat`

```json
{
  "id": "urn:zcap:root:https%3A%2F%2Fexample.com%2Ffoo",
  "controller": "did:key:example",
  "invocationTarget": "https://example.com/foo",
  "caveat": [
    {
      "type": "ValidWhileTrue",
      "uri": "https://example.com/status"
    }
  ]
}
```

### Root with unknown extension field

```json
{
  "id": "urn:zcap:root:https%3A%2F%2Fexample.com%2Ffoo",
  "controller": "did:key:example",
  "invocationTarget": "https://example.com/foo",
  "purpose": "not allowed on root"
}
```

All of these should fail validation because root zcaps must not have any fields beyond the exact root field set.

## Expected Behavior

For root capabilities, `ValidateCapabilityAsync` should reject:

- `Proof != null`
- `Expires != null`
- `ParentCapability != null`
- `AllowedAction != null` and/or non-empty
- `Caveat != null` and/or non-empty
- `AdditionalProperties != null && AdditionalProperties.Count > 0`

It should continue to accept root capabilities produced by `CreateRootCapabilityAsync`.

## Suggested Fix

Tighten the root branch in `CapabilityService.ValidateCapabilityAsync`:

```csharp
if (capability.Proof != null ||
    capability.Expires != null ||
    capability.ParentCapability != null ||
    capability.AllowedAction != null ||
    capability.Caveat != null ||
    capability.AdditionalProperties is { Count: > 0 })
{
    return Task.FromResult(false);
}
```

If backward compatibility requires tolerating empty arrays on in-memory objects, decide that explicitly and document it. For strict spec compliance, root `allowedAction` and `caveat` should be absent, not empty.

## Tests to Add

Add focused tests under `tests/ZcapLd.Core.Tests/Compliance/` or `tests/ZcapLd.Core.Tests/Services/CapabilityServiceTests.cs`:

1. `ValidateCapability_RootWithAllowedAction_ShouldReturnFalse`
2. `ValidateCapability_RootWithCaveat_ShouldReturnFalse`
3. `ValidateCapability_RootWithAdditionalProperty_ShouldReturnFalse`
4. `ValidateCapability_RootFromCreateRootCapabilityAsync_ShouldReturnTrue`
5. JSON deserialization test proving an unknown root field lands in `AdditionalProperties` and then fails validation.

## Acceptance Criteria

- Root creation still emits only `@context`, `id`, `controller`, and `invocationTarget`.
- Root validation rejects every field outside that exact set.
- Compliance tests cover forbidden root `allowedAction`, `caveat`, and extension fields.
- Full test suite passes.

## Related Context

This was found during a source-of-truth compliance re-check against the live W3C CCG draft. The current compliance matrix checks that generated roots omit optional fields, but it does not yet verify that externally supplied invalid root shapes are rejected.

moisesja:
## Summary

Fixes #49. Per the W3C CCG ZCAP-LD v0.3 draft, a root zcap contains exactly `@context`, `id`, `controller`, and `invocationTarget`, and **"A root zcap MUST NOT have any other fields."**

`CapabilityService.CreateRootCapabilityAsync` already emits this strict shape, but `ValidateCapabilityAsync` was too permissive: its root branch rejected only `Proof`, `Expires`, and `ParentCapability`. Externally-supplied roots (e.g. another stack's wire body) carrying `allowedAction`, `caveat`, or unknown extension fields — the latter captured into `AdditionalProperties` via `[JsonExtensionData]` — silently passed validation.

## Change

Tighten the `isRoot` branch in `ValidateCapabilityAsync` to also reject:

- `AllowedAction != null`
- `Caveat != null`
- `AdditionalProperties is { Count: > 0 }`

**Strict policy:** these must be absent (`null`), not empty arrays — the spec omits them on roots and `CreateRootCapabilityAsync` leaves them `null`, so no real producer emits empty arrays. Documented inline.

**Scope:** root capabilities only. Delegated zcaps legitimately carry these fields and are unchanged.

## Tests

- `CapabilityServiceTests`:
  - `ValidateCapability_RootWithAllowedAction_ShouldReturnFalse` (uses valid `"read"` so the SHOULD-05 action check passes first and the new root check is what fires)
  - `ValidateCapability_RootWithCaveat_ShouldReturnFalse`
  - `ValidateCapability_RootWithAdditionalProperty_ShouldReturnFalse` (deserializes a wire body → asserts the unknown field lands in `AdditionalProperties` → validation fails)
  - `ValidateCapability_RootDeserializedFromCleanJson_ShouldReturnTrue` (positive round-trip)
- `NormativeUnitComplianceTests`: `MUST-03b` covers the **validation** path (existing `MUST-03` covers the **creation** path).

## Verification

- `dotnet build ZcapLd.sln` → 0 errors
- Full suite → **356/356 pass**, no regressions


moisesja (commented):
I don't see any change log.

PR #78, opened 2026-06-02
fix(#58): normalize Proof.Capability so JSON-deserialized invocations verify

### LINKED ISSUE #58: [H1] VerifyInvocationAsync rejects all JSON-deserialized invocations (Proof.Capability as string)
## Summary

`VerifyInvocationAsync` reads `invocation.Proof.Capability` with an `as string` cast, but the property is typed `object?` and `System.Text.Json` deserializes it to a `JsonElement` — never a `string`. As a result the cast yields `null`, the proof/invocation field-consistency check fails, and **every invocation that arrives over the wire (received as JSON, then deserialized) is rejected**, even a perfectly valid one.

This is the canonical resource-server flow (receive JSON → deserialize → verify), so it is broken for real consumers. It only works in-process because `SigningService` assigns a CLR `string`.

## Location

- [`src/ZcapLd.Core/Services/VerificationService.cs:248`](src/ZcapLd.Core/Services/VerificationService.cs#L248) — the `as string` read
- [`src/ZcapLd.Core/Models/Proof.cs:66-68`](src/ZcapLd.Core/Models/Proof.cs#L66-L68) — `Capability` declared `object?`

```csharp
// VerificationService.cs ~248
if (!string.Equals(invocation.Proof.Capability as string, invocation.Capability, StringComparison.Ordinal) ||
    !string.Equals(invocation.Proof.CapabilityAction, invocation.CapabilityAction, StringComparison.Ordinal) ||
    !string.Equals(invocation.Proof.InvocationTarget, invocation.InvocationTarget, StringComparison.Ordinal))
    return false;   // line 251
```

After JSON round-trip, `invocation.Proof.Capability` is a `JsonElement` (ValueKind = String), so `as string` returns `null`, `string.Equals(null, invocation.Capability)` is `false`, and verification returns `false`.

## Reproduction

```csharp
var didProvider   = new InMemoryDidProvider();                 // test signer + resolver
var signing       = new SigningService(didProvider, didProvider);
var capabilities  = new CapabilityService(signing);
var verification  = new VerificationService(didProvider, new CaveatProcessor());

var aliceDid = "did:key:z6MkAlice";
didProvider.GenerateAndRegisterKeyPair(aliceDid);

    aliceDid, "https://example.com/resource", new[] { "read" });

var invocation = new Invocation
{
    Id               = $"urn:uuid:{Guid.NewGuid()}",
    Capability       = root.Id,
    CapabilityAction = "read",
    InvocationTarget = "https://example.com/resource",
};
invocation.Proof = await signing.SignInvocationAsync(invocation, aliceDid);

// 1) In-memory verify — PASSES (Proof.Capability is a CLR string here)
bool inMemory = await verification.VerifyInvocationAsync(invocation, root);   // true

// 2) Round-trip through JSON, as a resource server would receive it
var json     = JsonSerializer.Serialize(invocation, ZcapJsonOptions.Default);
var received = JsonSerializer.Deserialize<Invocation>(json, ZcapJsonOptions.Default);

// 3) Verify the deserialized invocation — FAILS
bool overTheWire = await verification.VerifyInvocationAsync(received!, root);  // false  ← BUG
```

`inMemory == true` but `overTheWire == false` for the same, valid, untampered invocation.

The existing test suite never round-trips an invocation through JSON before calling `VerifyInvocationAsync`, which is why this shipped.

## Impact

**High.** Any service that verifies invocations it received as serialized JSON (the normal resource-server / verifier deployment) rejects all of them. The in-process sign-then-verify path masks it.

## Recommended fix

Use the library's existing normalizer instead of the `as string` cast:

```csharp
var proofCapabilityId = invocation.Proof.Capability is { } cap ? TryExtractStringValue(cap) : null;
// TryExtractStringValue already handles boxed string and JsonElement(String) — VerificationService.cs:631-644
```

Stronger fix: attach a custom `JsonConverter` to the `capability` field so the polymorphic `object?` is normalized to a `string` on deserialize and never leaks to consumers. The same untyped-field pattern exists on `Capability.Context` and `Proof.CapabilityChain`, but those are read through helpers and are warts rather than bugs.

Add a regression test: **sign → serialize with `ZcapJsonOptions.Default` → deserialize → `VerifyInvocationAsync` must return `true`.**

---
**How found:** surfaced during `net-wallet-sdk` Phase 2 integration; independently confirmed by two adversarial reviewers (one reproduced the pass→fail end-to-end against the real library).
**Audit ref:** H1

moisesja:
## Summary

Fixes #58 (H1, high severity). `VerifyInvocationAsync` rejected **every** invocation received as JSON.

`Proof.Capability` is typed `object?`. After `System.Text.Json` deserialization it is a `JsonElement` (kind `String`), not a CLR `string`. The proof/invocation consistency check read it with an `as string` cast, which always yielded `null` → `string.Equals(null, invocation.Capability)` is `false` → verification returned `false`. This broke the canonical resource-server flow (receive JSON → deserialize → verify); it only ever passed in-process, where `SigningService` assigns a CLR `string`.

## Fix

Normalize `Proof.Capability` through the library's existing `TryExtractStringValue` helper (handles both a boxed `string` and a `JsonElement` of kind `String`) instead of the `as string` cast.

## Verification

- New regression test `VerificationServiceTests.VerifyInvocation_AfterJsonRoundTrip_ShouldReturnTrue`: sign → serialize with `ZcapJsonOptions.Default` → deserialize → verify must return `true`.
- Confirmed the test **fails** on `main` (reproduces the bug) and **passes** with the fix.
- Full suite green: 357 passed (356 baseline + 1 new).


moisesja (commented):
Solid fix — clear root-cause diagnosis, minimal change, good regression test.

**Fix** (`VerificationService.cs` ~line 276): Using the existing `TryExtractStringValue` helper to normalize `Proof.Capability` is exactly right. The `object?` typing of that field makes JSON round-trip normalization necessary, and reusing the in-library helper means no new code paths to maintain.

**Test** (`VerifyInvocation_AfterJsonRoundTrip_ShouldReturnTrue`): The structure is clean — verify in-process first (documents the previously-passing baseline), then round-trip through JSON to reproduce the bug, then assert the deserialized result verifies. The comment explaining the regression is useful context for future readers.

**One observation for a follow-up** (not blocking): `Proof.Capability` is the only `object?`-typed field compared in the consistency block; `CapabilityAction` and `InvocationTarget` are `string?` so they're fine. If future `object?` fields are added to `Proof` or `Invocation`, they'd hit the same pitfall — worth adding a short comment near the `object?` property declaration warning contributors to use `TryExtractStringValue` rather than `as string`.

357 tests green, changelog updated, scope is tight. LGTM.

---

PR #79, opened 2026-06-02
fix(#59): allow application-defined actions in ValidateCapabilityAsync

### LINKED ISSUE #59: [M1] ValidateCapabilityAsync rejects any action other than read/write
## Summary

`ValidateCapabilityAsync` rejects any capability whose `allowedAction` contains a value other than `read` or `write`. ZCAP-LD treats actions as an application-defined vocabulary — `read`/`write` are examples, not a closed allow-list — so this declares spec-conformant capabilities invalid, including ones this very library mints and cryptographically verifies.

## Location

[`src/ZcapLd.Core/Services/CapabilityService.cs:163-175`](src/ZcapLd.Core/Services/CapabilityService.cs#L163-L175)

```csharp
// COMPLIANCE FIX: SHOULD-05 - Validate action names (should be read/write)
if (capability.AllowedAction != null && capability.AllowedAction.Length > 0)
{
    var validActions = new[] { "read", "write" };
    foreach (var action in capability.AllowedAction)
    {
        if (!validActions.Contains(action, StringComparer.OrdinalIgnoreCase))
        {
            return Task.FromResult(false);   // ← rejects "delete", "sign", "admin", "share", …
        }
    }
}
```

## Why this is wrong

- The repo's own spec notes treat actions as application-defined: `docs/ZCAP-LD-SPECIFICATION-REQUIREMENTS.md:461,775-778` ("Application-specific values allowed", example `Drive`).
- The library is internally inconsistent: `CreateRootCapabilityAsync` / `DelegateCapabilityAsync` mint arbitrary actions; `ValidateAttenuation` and `VerifyInvocationAsync` only do subset/membership checks. So a capability the library **produces and verifies** is declared invalid by its own `ValidateCapabilityAsync`.
- The library's own integration tests delegate `delete`/`admin`/`share` (`EndToEndTests.cs:112,571`; `RdfcEndToEndTests.cs:118,163`).
- The treatment is inconsistent with adjacent SHOULD checks (root-id at lines 190-193, urn:uuid at 222-226) that only warn rather than fail.

## Reproduction

```csharp
var signing      = new SigningService(didProvider, didProvider);
var capabilities = new CapabilityService(signing);

    aliceDid, "https://example.com/doc", new[] { "delete" });

var bob = await capabilities.DelegateCapabilityAsync(
    root, bobDid, new[] { "delete" }, expires: DateTime.UtcNow.AddDays(1));

bool valid = await capabilities.ValidateCapabilityAsync(bob);   // false  ← despite verifying fine
bool chainOk = await new VerificationService(didProvider, new CaveatProcessor())
                        .VerifyCapabilityChainAsync(bob);        // true
```

`ValidateCapabilityAsync` says invalid; `VerifyCapabilityChainAsync` says valid. Same capability.

## Impact

**Medium** (fails closed — over-rejects, never grants). Any capability modeling a real action other than read/write is reported invalid even though it is spec-conformant and cryptographically valid.

## Recommended fix

Remove the closed allow-list block (lines 163-175). The real constraints — attenuation (child ⊆ parent) and invocation-time membership in `allowedAction` — are already enforced elsewhere. If vocabulary restriction is genuinely desired, make it an injected, opt-in policy that defaults to "any non-empty string." Update the `SHOULD-05` test (`NormativeUnitComplianceTests.cs:297-317`) to assert custom actions validate.

---
**How found:** adversarial multi-agent audit (found independently by the API-contract and spec-compliance reviewers; both confirmed real).
**Audit ref:** M1

moisesja:
## Summary

Fixes #59 (M1). `ValidateCapabilityAsync` rejected any capability whose `allowedAction` contained a value other than `read`/`write`.

ZCAP-LD treats actions as an **application-defined vocabulary** — `read`/`write` are illustrative, not a closed allow-list. The hard-coded check declared spec-conformant capabilities invalid, including ones this library itself mints (`delete`, `admin`, `share`) and cryptographically verifies via `VerifyCapabilityChainAsync`. It was internally inconsistent: `DelegateCapabilityAsync` mints arbitrary actions, and attenuation / invocation-time checks only do subset/membership validation.

## Fix

Remove the closed allow-list block from `ValidateCapabilityAsync`. The real constraints remain enforced elsewhere:
- **attenuation** (child ⊆ parent) at delegation and chain verification,
- **membership** in `allowedAction` at invocation time.

A deployment wanting vocabulary restriction should layer it as an injected policy (out of scope here).

## Verification

- `SHOULD-05` compliance test (`NormativeUnitComplianceTests`) flipped to assert a custom `admin` action **validates** (was asserting it failed).
- Full suite green: 357 passed.

> Note: cumulative branch — opened off the #58 branch, so this PR's diff includes #58 until that merges.


moisesja (commented):
## Review: PR #79 — allow application-defined actions in ValidateCapabilityAsync

Both fixes are spec-correct, well-motivated, and properly tested. This looks good to merge.

---

### `CapabilityService.cs` — allow-list removal (Issue #59)

Correct call. ZCAP-LD v0.3 explicitly describes `read`/`write` as illustrative examples; `allowedAction` is application-defined vocabulary. The hard-coded check created an internal contradiction: `DelegateCapabilityAsync` happily mints arbitrary actions and `VerifyCapabilityChainAsync` cryptographically verifies them — only `ValidateCapabilityAsync` rejected them. The real enforcement (attenuation child ⊆ parent, membership at invocation time) is unaffected by this removal. Leaving vocabulary restriction to an injected policy is the right boundary.

### `VerificationService.cs` — `JsonElement` normalization (Issue #58)

Root cause is clear: `Proof.Capability` is `object?`, and `System.Text.Json` deserializes JSON strings into `JsonElement`, not CLR `string`. The `as string` cast always produced `null` on the deserialized path, silently rejecting every wire invocation while in-process tests continued to pass. Using `TryExtractStringValue` (already present in the codebase) is the right fix — it handles both boxed `string` and `JsonElement` of kind `String` in one place.

### Tests

- `SHOULD-05` assertion flip is semantically correct (a custom `admin` action is spec-conformant).
- `VerifyInvocation_AfterJsonRoundTrip_ShouldReturnTrue` is a well-targeted regression guard — it explicitly exercises the JSON serialize → deserialize → verify path that was broken, making the failure mode visible in the test name itself.

### One note: cumulative branch

The diff includes both #58 and #59 fixes (acknowledged in the PR description). If #58 is still open as a separate PR, consider rebasing this branch onto `main` once #58 merges so the commit history stays scoped. No blocker — just a hygiene note.

No concerns overall. 357 tests passing, CHANGELOG updated, reasoning is sound.

---

PR #80, opened 2026-06-02
fix(#60): add authorizing revocation overload; mark string path as no-authz

### LINKED ISSUE #60: [M3] RevokeCapabilityAsync performs no revoker authorization
## Summary

`RevokeCapabilityAsync(capabilityId, revokerDid)` performs **no authorization**: it validates only that the two strings are non-empty, stores a revocation record, and unconditionally returns `true`. `revokerDid` is recorded only as audit attribution (`RevocationRecord.RevokedBy`) and is never checked against the capability's controller or delegation chain. Because a stored revocation immediately flips `VerifyCapabilityChainAsync` / `VerifyInvocationAsync` to `false`, **any caller can revoke anyone's capability** — an unauthenticated denial-of-capability vector. The boolean return is also vacuous (never `false` on the success path).

## Location

[`src/ZcapLd.Core/Services/VerificationService.cs:658-689`](src/ZcapLd.Core/Services/VerificationService.cs#L658-L689) (interface doc at `IVerificationService.cs:53-57`)

```csharp
public Task<bool> RevokeCapabilityAsync(string capabilityId, string revokerDid)
{
    if (string.IsNullOrEmpty(capabilityId)) throw new ArgumentException(...);
    if (string.IsNullOrEmpty(revokerDid))   throw new ArgumentException(...);
    return RevokeCapabilityCoreAsync(capabilityId, revokerDid);
}

private async Task<bool> RevokeCapabilityCoreAsync(string capabilityId, string revokerDid)
{
    await _revocationService.RevokeAsync(new RevocationRequest
    {
        CapabilityId = capabilityId,
        RevokedBy    = revokerDid     // stored as attribution only; never authorized
    });
    return true;                      // never false on success
}
```

The ASP.NET endpoint (`ZcapRevocationEndpointRouteBuilderExtensions.cs:46-74`) likewise accepts a bare, unsigned `revokerDid` with no proof.

## Reproduction

```csharp
var verification = new VerificationService(didProvider, new CaveatProcessor());

// bobsCap was delegated by Alice to Bob and verifies fine
(await verification.VerifyCapabilityChainAsync(bobsCap)).Should().BeTrue();

// Mallory — who controls nothing in this chain — revokes it:
await verification.RevokeCapabilityAsync(bobsCap.Id, "did:key:z6MkMalloryHasNoAuthority");

// Now it no longer verifies, for everyone:
(await verification.VerifyCapabilityChainAsync(bobsCap)).Should().BeFalse();
```

## Impact

**Medium.** The method lives on `IVerificationService`, takes a `revokerDid` that *implies* an authority check, and grants effective revocation to any caller. A consumer reading the signature reasonably assumes an unauthorized revoker yields `false`; instead it succeeds.

## Recommended fix

1. **Authorization** — require the revoker to be the capability's controller or an up-chain delegator (reuse `IsControllerAuthorized` / chain logic), ideally via a *signed* revocation invocation rather than a bare string; update the ASP.NET endpoint in lockstep. If host-side authorization is the deliberate design, rename/document the method to state explicitly that it performs **no** authorization.
2. **Boolean contract** — either make `false` reachable (e.g. on an authorization failure) or change the return to `Task`, and fix the XML doc that currently claims "True if revocation was successful."

---
**How found:** adversarial multi-agent audit (security dimension; one reviewer rated medium emphasizing unauthenticated revocation, one low framing it as host responsibility — kept at medium because the method signature implies authority and the effect is immediate).
**Audit ref:** M3

moisesja:
## Summary

Fixes #60 (M3, security). `RevokeCapabilityAsync(string capabilityId, string revokerDid)` performed **no authorization** — it stored a revocation and unconditionally returned `true`. Since a stored revocation immediately flips `VerifyCapabilityChainAsync`/`VerifyInvocationAsync` to `false`, any caller could revoke anyone's capability (unauthenticated denial-of-capability), and the `bool` return was vacuous.

## Fix

A bare `capabilityId` + `revokerDid` cannot be authorized in-method (only the ID is available, and a bare DID string is not proof of control), so:

1. **Honest contract on the string overload** — documented explicitly as performing **no** authorization (the `revokerDid` is audit attribution only; the host must authorize before calling). Corrected the misleading "True if revocation was successful" XML doc.
2. **New authorizing overload** `RevokeCapabilityAsync(Capability, string)` — verifies the revoker controls the capability or any ancestor in its delegation chain (an up-chain delegator), and returns `false` **recording nothing** when unauthorized or the chain can't be verified. `revokerDid` is a DID the host has already *authenticated*; the library performs authorization, not authentication. This makes a `false` return reachable and meaningful.

The `ZcapLd.AspNetCore` revocation endpoint still accepts a bare `revokerDid` (it calls `IRevocationService.RevokeAsync` directly); binding it to a *signed* revocation invocation is a larger feature tracked separately.

## Verification

New tests: authorized up-chain controller revokes (`true` + revoked); unauthorized revoker refused (`false`, not revoked, still verifies); own-controller self-revocation succeeds. Full suite green: 360 passed (357 + 3).

> Cumulative branch — diff includes #58 and #59 until they merge.


moisesja (commented):
Good security fix overall — the diagnosis of the original gap (string overload stores revocations unconditionally) is exactly right, and the approach of adding an authorizing overload while honestly documenting the string overload is sound. Three items worth addressing before merging:

---

## 1. `IsRevokerAuthorizedAsync` uses structural chain traversal, not cryptographic verification (security-significant)

`BuildCapabilityChainAsync` validates structural integrity (IDs match, no cycles, root invariants) but **does not verify the proof signatures** on delegation links. This means a caller who passes a crafted `Capability` CLR object — e.g., one deserialized from an attacker-controlled request body — with a tampered `controller` field and a structurally plausible (but cryptographically unsigned) chain could pass the authorization check, causing a legitimate capability to be revoked. The very denial-of-capability attack this PR aims to prevent remains possible through a different vector.

Concrete attack sketch:
1. Mallory sends a revocation HTTP request with the victim's capability ID and a crafted capability body where `controller` is Mallory's DID and the embedded root in `capabilityChain` also names Mallory as controller.
2. The host deserializes the capability and calls `RevokeCapabilityAsync(deserializedCapability, malloryDid)`.
3. `BuildCapabilityChainAsync` builds the chain structurally (no signature check), finds Mallory's DID in a controller field, returns authorized.
4. The victim capability is revoked.

**Suggested fix** — verify the chain cryptographically before trusting its controller fields:

```csharp
private async Task<bool> IsRevokerAuthorizedAsync(Capability capability, string revokerDid)
{
    try
    {
        // Verify signatures first so controller fields can be trusted
        if (!await VerifyCapabilityChainAsync(capability))
            return false;

        var chain = await BuildCapabilityChainAsync(capability);
        return chain.Any(link =>
            link.Controller is not null &&
            link.Controller.ContainsVerificationMethod(revokerDid));
    }
    {
        return false;
    }
}
```

If the intention is that callers verify the chain themselves before calling this method, that assumption needs to be stated explicitly in the XML doc — and ideally the method should throw `InvalidOperationException` (or return false) when chain verification fails, rather than proceeding to the authorization check on unverified data.

---

## 2. String overload remains a footgun despite honest docs

Marking the existing `RevokeCapabilityAsync(string, string)` as "host must authorize before calling" is the right kind of honesty, but for a security-critical, irreversible operation, consider adding an `[Obsolete]` attribute to surface the risk at compile time:

```csharp
[Obsolete("This overload performs no authorization. Prefer RevokeCapabilityAsync(Capability, string), " +
          "which verifies the revoker against the delegation chain. Call this overload only when " +
          "the host has already enforced authorization externally.", error: false)]
Task<bool> RevokeCapabilityAsync(string capabilityId, string revokerDid);
```

Non-error `[Obsolete]` keeps the existing `AspNetCore` endpoint compiling (tracked separately) while making the risk visible in any IDE or `dotnet build` output.

---

## 3. Missing argument-guard tests for the new overload

The implementation throws on null/empty inputs, but there are no tests for those paths. Consider adding:

```csharp
[Fact]
public async Task RevokeCapability_NullCapability_ThrowsArgumentNullException() { ... }

[Fact]
public async Task RevokeCapability_EmptyRevokerDid_ThrowsArgumentException() { ... }
```

---

## Minor notes

- **Cumulative branch**: the diff bundles the `allowedAction` vocabulary fix (Issue #59) alongside Issue #60. Calling that out in the title or squashing cleanly per issue would make the commit history easier to trace.
- `ContainsVerificationMethod` (split on `#`, match bare DID or full VM URI) is the correct logic for authorization — once the chain is cryptographically verified.
- The three new revocation tests clearly cover authorized (up-chain delegator), unauthorized (Mallory), and self-revocation cases — solid coverage for the happy/sad paths.


---

Issue #73, opened 2026-06-02
[L8] Verifier-side 3-month expiration SHOULD is absent


moisesja:
## Summary

The W3C ZCAP-LD verifier-side SHOULD — "a verifier SHOULD ensure an invoked delegated zcap does not have an expiration more than three months in the future" — is **not implemented on the verify path**. The MUST "not expired" check is correctly present, but the 3-month ceiling lives (wrongly) on the create path as a hard throw (audit ref **M4**). This issue tracks adding the check where it belongs.

## Location

[`src/ZcapLd.Core/Services/VerificationService.cs:367-370`](src/ZcapLd.Core/Services/VerificationService.cs#L367-L370) — only the basic expiry (`childExpiresAt < UtcNow`) check exists; no 3-month-ahead ceiling.

```csharp
// present: MUST not-expired
var childExpiresAt = child.ExpiresAt;
if (childExpiresAt.HasValue && childExpiresAt.Value < DateTime.UtcNow)
    return false;
// missing: verifier-side SHOULD that expiry is not > 3 months in the future at invocation time
```

## Impact

**Low.** In isolation, omitting a SHOULD on verify is defensible. This entry exists to pair with **M4**: the fix should *add* this check on the verifier (measured at invocation time, behind a policy flag because it is a SHOULD) **while removing** the create-time hard throw.

## Recommended fix

Treat as the same work item as **M4**:
1. Remove the create-time throw (`CapabilityService.cs:270-280`).
2. Add the 3-month ceiling here, measured at verification time, gated behind a verifier policy flag, with `MaxExpirationMonths` made configurable.
3. Fix the misnamed compliance test (`NormativeIntegrationComplianceTests.cs:347`) to exercise the verifier rather than `DelegateCapabilityAsync`.

---
**Affected version:** `ZcapLd.Core` 2.1.1 (source verified at tag `core-v2.1.1`)
**How found:** adversarial multi-agent audit (spec-compliance). Same work item as **M4**.
**Audit ref:** L8

moisesja:
Same work item as #61 (M4): remove the create-time hard throw there and add the ceiling here, measured at verification time behind a policy flag.

PR #81, opened 2026-06-02
fix(#61): stop hard-throwing on long delegated expirations at create time

### LINKED ISSUE #61: [M4] 3-month expiration cap is a hard throw at delegation time (should be a verifier-side SHOULD)
## Summary

The W3C ZCAP-LD 3-month expiration ceiling is implemented as a **hard `throw` at delegation (create) time**, but the spec makes it a **verifier-side SHOULD measured at invocation time**. Three problems: (1) a SHOULD enforced as a hard throw; (2) wrong actor/time — it blocks even *constructing* a longer-lived delegation; (3) the ceiling (`MaxExpirationMonths = 3`) is a non-configurable `const` with no opt-out. Meanwhile the verifier never enforces it (see companion issue, audit ref **L8**).

## Location

[`src/ZcapLd.Core/Services/CapabilityService.cs:270-280`](src/ZcapLd.Core/Services/CapabilityService.cs#L270-L280) (const at line 15); absent on the verify side at `VerificationService.cs:367-370`.

```csharp
// CapabilityService.cs — inside ValidateAttenuation, runs during DelegateCapabilityAsync
if (expires.HasValue)
{
    var maximumExpirationDate = DateTime.UtcNow.AddMonths(MaxExpirationMonths); // const 3
    if (expires.Value > maximumExpirationDate)
    {
        throw new InvalidOperationException(
            $"Capability expiration exceeds recommended {MaxExpirationMonths}-month limit. ...");
    }
}
```

The cited spec text (`docs/ZCAP-LD-SPECIFICATION-REQUIREMENTS.md:720`): *"A verifier SHOULD ensure that an invoked delegated zcap does not have an expiration date-time more than three months in the future"* — a SHOULD, on the **verifier**, at **invocation** time.

## Reproduction

```csharp
var root = await capabilities.CreateRootCapabilityAsync(aliceDid, "https://example.com/doc", new[]{"read"});

// Delegate valid for a year — well within the parent's authority (root has no expiry):
await capabilities.DelegateCapabilityAsync(
    root, bobDid, new[] { "read" }, expires: DateTime.UtcNow.AddYears(1));
// → throws InvalidOperationException("…exceeds recommended 3-month limit")
```

There is no way to opt out: `MaxExpirationMonths` is a `const` and the check is unconditional.

## Impact

**Medium.** Consumers cannot create legitimately long-lived delegations even when the parent permits it, and the actual spec protection (bounding verifier-side stored-revocation burden) is absent. The compliance test `NormativeIntegrationComplianceTests.cs:347` is named a *verifier* test but drives `DelegateCapabilityAsync`, green-lighting a verifier requirement the verifier never implements.

## Recommended fix

1. Remove/soften the throw at `CapabilityService.cs:270-280` (warning at most).
2. Add the ceiling to `VerifyCapabilityChainAsync` / `VerifyInvocationAsync` **measured at verification time**, gated behind a verifier policy flag (it is a SHOULD). This is the companion issue **L8**.
3. Make `MaxExpirationMonths` configurable.
4. Fix the misnamed test to exercise the verifier.

---
**How found:** adversarial multi-agent audit (spec-compliance). Companion: **L8** (verify-side SHOULD absent) — same work item.
**Audit ref:** M4

moisesja:
## Summary

Fixes #61 (M4, spec-compliance). The W3C ZCAP-LD 3-month expiration ceiling is a **verifier-side SHOULD measured at invocation time**, but `CapabilityService.DelegateCapabilityAsync` enforced it as a hard `throw` at create time inside `ValidateAttenuation`. That (1) turned a SHOULD into a hard failure, (2) used the wrong actor/time — blocking even *constructing* a legitimately long-lived delegation the parent permits — and (3) had no opt-out (`MaxExpirationMonths` was a non-configurable `const`).

## Fix

- Remove the create-time 3-month throw and the now-unused `MaxExpirationMonths` const from `CapabilityService`. Attenuation (child ≤ parent expiry) is still enforced.
- The ceiling moves to the **verify path behind a policy flag** in the companion issue #73 (cumulative successor of this branch).
- Repurpose the misnamed `SHOULD-04` compliance test (it was named a *verifier* test but drove `DelegateCapabilityAsync`) to assert long-lived delegation succeeds at create time; update the compliance test matrix.

## Verification

Full suite green: 360 passed.

> Cumulative branch — diff includes #58–#60 until they merge.


moisesja (commented):
Good set of spec-compliance fixes. The reasoning in the PR description and inline comments is clear, all three changes move in the right direction, and 360 tests green is a solid signal. A few observations below.

---

**`IsRevokerAuthorizedAsync` — authorization semantics worth double-checking**

```csharp
return chain.Any(link =>
    link.Controller is not null &&
    link.Controller.ContainsVerificationMethod(revokerDid));
```

`revokerDid` is an *authenticated* DID (e.g. `did:key:z6MkRevokeRoot`), but `ContainsVerificationMethod` tests whether the controller's DID Document lists that value as a verification method. For a `did:key` controller that is also the revoker this is fine, but for cases where the controller and the signing key are different DIDs (e.g. controller = `did:web:issuer.example` and revokerDid = a specific key fragment `did:web:issuer.example#key-1`) the check may return `false` when it should return `true`. The tests all use the same DID as both controller and revoker, so this edge case isn't exercised. Worth at least a TODO or a dedicated test.

---

**SHOULD-04 assertion is minimal**

```csharp
delegated.Expires.Should().NotBeNull();
```

The test proves no exception was thrown (the primary goal of the issue), but it doesn't assert the value is actually the requested 6-month expiration. A tighter assertion — `delegated.Expires.Should().BeCloseTo(DateTime.UtcNow.AddMonths(6), precision: TimeSpan.FromSeconds(5))` — would catch a silent truncation in `DelegateCapabilityAsync`. Not blocking, but it's a gap worth closing.

---

**Acknowledged gap: no expiration ceiling enforcement until #73**

The PR description explicitly calls this out, and the decision is defensible. Just flagging it clearly for reviewers: once this merges, there is no enforcement of the 3-month ceiling anywhere in the library until #73 ships. If there's any chance #73 is delayed, a release note or warning log at `DelegateCapabilityAsync` for long expirations might be worth considering.

---

**Cumulative branch note**

The diff includes changes for #58–#61 bundled together. The CHANGELOG is thorough and each change is independently sound, but reviewers who approved earlier PRs in the stack will want to re-verify their assumptions here if #58–#60 haven't merged yet.

---

Overall this is a well-reasoned set of fixes. The `IsRevokerAuthorizedAsync` key-fragment edge case and the weak SHOULD-04 assertion are the only items I'd flag before merge.

---

PR #82, opened 2026-06-02
fix(#62): enable replay protection by default in convenience constructors

### LINKED ISSUE #62: [M5] Replay protection disabled by default (NullNonceStore in convenience constructors)
## Summary

The convenience `VerificationService` constructors disable replay protection by default: the 2-, 3-, and 4-argument constructors all funnel to `NullNonceStore.Instance`, whose `TryMarkAsUsedAsync` always returns `false`, making the nonce/replay step a no-op. The documented backward-compatible path `new VerificationService(resolver, caveatProcessor)` therefore gives **zero** replay protection — a captured valid invocation can be replayed indefinitely. This is the opposite of the project's own stated security posture.

## Location

[`src/ZcapLd.Core/Services/VerificationService.cs:64-70`](src/ZcapLd.Core/Services/VerificationService.cs#L64-L70) (replay step neutered at lines 308-311)

```csharp
// 2-arg ctor → ... → this(..., NullNonceStore.Instance)   (line 70)
public VerificationService(IDidResolver didResolver, ICaveatProcessor caveatProcessor)
    : this(didResolver, caveatProcessor, CreateDefaultSuiteProvider(),
           new RevocationService(new InMemoryRevocationStore())) { }
// → chains to the ctor that passes NullNonceStore.Instance
```

```csharp
// Step 8 becomes a no-op with NullNonceStore (lines 308-311)
var nonceExpiry = DateTime.UtcNow.Add(_nonceWindow);
if (await _nonceStore.TryMarkAsUsedAsync(invocation.Id, nonceExpiry))   // always false
    return false;
```

**Evidence it is a defect, not intent:** `docs/SECURITY-FIXES-SUMMARY.md:177-178` states the intended posture is the *opposite* — "InMemoryNonceStore (default)", "NullNonceStore available for opt-out." `README.md:20` advertises replay protection as built-in, while `README.md:48`'s Quick Start uses the insecure 2-arg constructor. The maintainers' own `VerificationServiceReplayTests.cs:110-121` demonstrate the no-protection behavior.

## Reproduction

```csharp
// The documented, most-used construction path:
var verification = new VerificationService(didProvider, new CaveatProcessor());

var invocation = /* a valid, signed invocation with a fixed Id */;

bool first  = await verification.VerifyInvocationAsync(invocation, capability);  // true
bool replay = await verification.VerifyInvocationAsync(invocation, capability);  // true  ← replayed
bool again  = await verification.VerifyInvocationAsync(invocation, capability);  // true  ← and again
```

With `NullNonceStore` the same captured invocation verifies an unlimited number of times. (Even the explicit `InMemoryNonceStore` is process-local, so multi-instance deployments need a shared store — worth documenting.)

## Impact

**Medium.** Insecure-by-default footgun; safe and unsafe call sites are indistinguishable. The ASP.NET DI path (`AddZcapServices`) correctly wires `InMemoryNonceStore`, so only direct `new` / 2-3-4-arg construction is affected — but that is the path the README Quick Start shows.

## Recommended fix

- Default the convenience constructors to `new InMemoryNonceStore()`.
- Make `NullNonceStore` reachable only via the explicit `INonceStore` constructor, so opting out is deliberate.
- Update the XML doc and the README Quick Start.
- Add a regression test asserting a replay is rejected with the 2-arg constructor.
- Document that `InMemoryNonceStore` is process-local (supply a shared store across instances for multi-node verifiers).

---
**How found:** adversarial multi-agent audit (security). Both reviewers confirmed; severity downgraded from "high" to medium because exploitation requires a fixed `invocation.Id` and even `InMemoryNonceStore` is process-local.
**Audit ref:** M5

moisesja:
## Summary

Fixes #62 (M5, security). The 2-, 3-, and 4-argument `VerificationService` constructors funnelled to `NullNonceStore.Instance`, whose `TryMarkAsUsedAsync` always reports "not seen" — so the documented `new VerificationService(resolver, caveatProcessor)` path (the README Quick Start) gave **zero** replay protection. A captured valid invocation could be replayed indefinitely. This was the opposite of the project's stated posture (`SECURITY-FIXES-SUMMARY.md`: "InMemoryNonceStore (default)").

## Fix

- The convenience constructors now default to `new InMemoryNonceStore()` (all chain through the 4-arg constructor, fixed in one place).
- `NullNonceStore` is reachable **only** by deliberately passing it to the explicit `INonceStore` constructor — opting out is now intentional.
- Documented that `InMemoryNonceStore` is process-local (use a shared store for multi-node verifiers) on the constructor and in the README Quick Start.

## Verification

- New `VerificationServiceReplayTests.VerifyInvocation_TwoArgConstructor_RejectsReplayByDefault` asserts the Quick Start path rejects a replay.
- Two existing tests reused one invocation id across multiple verifies (a round-trip control; the usage-count workflow). Enabling replay protection correctly broke them; both now use **distinct invocations per request** — the realistic flow, with the caveat (not replay) gating the workflow.
- Full suite green: 361 passed.

> Cumulative branch — diff includes #58–#61 until they merge.


moisesja (commented):
Good set of fixes — the core changes are all well-justified and the test coverage is solid. A few observations worth flagging before merge.

---

**The replay protection fix (Issue #62) is correct.**

The one-line change (`NullNonceStore.Instance` → `new InMemoryNonceStore()`) in the 4-arg constructor is exactly right, and chaining all convenience constructors through it keeps the fix in one place. The new regression test in `VerificationServiceReplayTests` covers the exact documented Quick Start path. The README and constructor XML docs are clear about the process-local limitation.

---

**Issue #60 — authorization logic worth a quick second look.**

`IsRevokerAuthorizedAsync` checks `link.Controller.ContainsVerificationMethod(revokerDid)`. Please confirm that `ContainsVerificationMethod` matches the revoker as the *controller* of the chain link, not merely as a key listed anywhere in the DID document. If it scans the `verificationMethod` array of an upstream controller's DID document, a DID that is a *key* of a controller (but not the controller subject itself) could revoke — which is probably not the intended semantics. The three new tests all use a controller DID that is also the subject, so they wouldn't catch this edge case.

---

**Issue #59 — SHOULD-05 test comment is misleading.**

```csharp
// The root carries no allowedAction (unrestricted), so delegating a custom "admin"
// action is within the parent's authority...
var root = await fixture.CapabilityService.CreateRootCapabilityAsync(
    parentDid,
    "https://example.com/resources",
    new[] { "read", "write" });   // ← explicitly passes ["read", "write"]
```

If `CreateRootCapabilityAsync` correctly strips `allowedAction` from root zcaps (as required by the spec — root zcaps MUST NOT carry the field), the comment is accurate but the call site is confusing: a reader sees "admin" ⊄ {"read", "write"} and wonders why attenuation isn't enforced. Consider asserting `root.AllowedAction` is null/empty, or passing `null`/`Array.Empty<string>()` to make the intent explicit.

---

**SHOULD-04 compliance matrix mapping.**

The matrix now maps SHOULD-04 to create-time delegation behaviour, but the spec normative text is about the *verifier*. The updated test name is accurate for what is tested, but a brief note in the matrix ("verifier-side enforcement tracked in #73") would prevent the row from looking like a full SHOULD-04 coverage claim.

---

**Scope note.**

The branch is cumulative across #58–#62 and the PR title names only #62. That's fine given the branching strategy, but a future `git revert` would undo all five fixes at once — worth noting for awareness.

---

Overall the changes are correct, well-tested (361 passing), and thoroughly documented. The SHOULD-05 comment and SHOULD-04 matrix note are minor enough not to block.

---

moisesja:
Addressed the review in 25a2c84.

**#59 — SHOULD-05 test comment (fixed).** `CreateRootCapabilityAsync` silently drops `allowedAction` on roots (it's intentionally unset — root zcaps grant unrestricted authority), so the old `new[] { "read", "write" }` call site read like an attenuation allow-list and made the "admin" delegation look unexplained. The test now passes `allowedActions: null` and asserts `root.AllowedAction` is null/empty, so the "root is unrestricted" premise is explicit and we're testing the application-vocabulary rule rather than attenuation.

**SHOULD-04 — compliance matrix note (fixed).** The matrix row now reads "create-time behaviour only; the spec's verifier-side 3-month expiration ceiling is tracked in #73 and not yet enforced," so it no longer looks like a full SHOULD-04 coverage claim.

**#60 — authorization logic (confirmed, no change needed).** `ContainsVerificationMethod` does **not** scan any DID document — it's a pure string-level match: the chain link's `controller` must equal either the revoker's bare DID (`vm.Split('#')[0]`) or the full verification-method URI. A key belonging to a *different* DID — even one a controller's DID document would authorize — is **not** matched (fail-closed). This is documented by the TODO in `VerificationService.IsRevokerAuthorizedAsync` and locked by `ControllerSetTests.ContainsVerificationMethod_DoesNotResolveCrossDidController`. Cross-DID key authorization via DID-document resolution is future work tracked by that TODO. So the edge case the three new revocation tests don't exercise is the *unsupported* (and intentionally rejected) direction — the conservative behaviour is already covered at the unit level.

**Scope note — acknowledged.** Noted that a `git revert` of this branch would undo #58–#62 together; no action.

All 365 tests pass.

Issue #69, opened 2026-06-02
[L4] VerifyCapabilityProofAsync does not enforce attenuation


moisesja:
## Summary

`VerifyCapabilityProofAsync` verifies a delegation's signature and parent-controller authorization, but does **not** enforce attenuation (child must be no broader than parent) or expiry. A re-signed child that *expands* authority beyond its parent is accepted by `VerifyCapabilityProofAsync`, while `VerifyCapabilityChainAsync` correctly rejects it. The risk is a consumer mistaking the proof method for a full authorization check — the name and doc ("Verifies a capability's cryptographic proof") invite exactly that.

## Location

[`src/ZcapLd.Core/Services/VerificationService.cs:141-144`](src/ZcapLd.Core/Services/VerificationService.cs#L141-L144) delegates to `VerifyDelegationProofAsync` with no `ValidateAttenuation` call; attenuation is only enforced in the chain verifier at `:364`.

## Reproduction

```csharp
// parent allows only ["read"]; forge a child that grants ["read","write"], re-signed by parent controller
var overbroadChild = /* child with AllowedAction = ["read","write"], valid delegation proof */;

bool proof = await verification.VerifyCapabilityProofAsync(overbroadChild);   // true   ← attenuation not checked
bool chain = await verification.VerifyCapabilityChainAsync(overbroadChild);   // false  ← correct
```

## Impact

**Low.** Not exploitable in-tree: the real access path `VerifyInvocationAsync` calls `VerifyCapabilityChainAsync` first (line 236). The hazard is a consumer using `VerifyCapabilityProofAsync` as an "is this authorized?" gate. (Same method, same misuse risk, as the revocation asymmetry in audit ref **M6**.)

## Recommended fix

Tighten the `IVerificationService` XML doc to state this method verifies ONLY single-link proof soundness + parent-controller authorization and is NOT an authorization gate — directing callers to `VerifyCapabilityChainAsync` / `VerifyInvocationAsync`. Optionally rename (e.g. `VerifyCapabilityProofSignatureAsync`) or have it additionally enforce attenuation/expiry against the embedded parent.

---
**Affected version:** `ZcapLd.Core` 2.1.1 (source verified at tag `core-v2.1.1`)
**How found:** adversarial multi-agent audit. Related: **M6** (revocation asymmetry in the same method).
**Audit ref:** L4

moisesja:
Related: #63 (M6) — ancestor revocation is also ignored in this same method.

PR #83, opened 2026-06-02
fix(#63): honour ancestor revocation in VerifyCapabilityProofAsync

### LINKED ISSUE #63: [M6] VerifyCapabilityProofAsync ignores ancestor revocation
## Summary

`VerifyCapabilityProofAsync` checks the **leaf's** revocation but ignores the **immediate parent's** revocation — even though the parent capability is right there in `Proof.CapabilityChain[^1]`. This is an incoherence with `VerifyCapabilityChainAsync`, which correctly checks revocation for every capability in the chain. A consumer who treats `VerifyCapabilityProofAsync` as "is this valid?" can accept a capability whose ancestor has been revoked.

## Location

[`src/ZcapLd.Core/Services/VerificationService.cs:126`](src/ZcapLd.Core/Services/VerificationService.cs#L126) (leaf-only check) vs the all-chain check at [`:342-348`](src/ZcapLd.Core/Services/VerificationService.cs#L342-L348)

```csharp
// VerifyCapabilityProofAsync ~126 — only the leaf id is checked
if (await IsCapabilityRevokedAsync(capability.Id))
    return false;
// ... then VerifyDelegationProofAsync verifies the signature, but never checks
//     the embedded parent (CapabilityChain[^1]) for revocation.
```

```csharp
// VerifyCapabilityChainAsync ~342-348 — correct: walks the whole chain
foreach (var chainCapability in chain)
    if (await IsCapabilityRevokedAsync(chainCapability.Id))
        return false;
```

## Reproduction

```csharp
// root → mid → leaf
await verification.RevokeCapabilityAsync(mid.Id, aliceDid);

bool proof = await verification.VerifyCapabilityProofAsync(leaf);    // true   ← ignores revoked parent
bool chain = await verification.VerifyCapabilityChainAsync(leaf);    // false  ← correct
```

## Impact

**Low–medium (leaning low).** The real enforcement path (`VerifyInvocationAsync` → `VerifyCapabilityChainAsync`) does check all-chain revocation, and no in-tree code uses `VerifyCapabilityProofAsync` as a sufficiency gate. The risk is a consumer mistaking the single-proof method for a full authorization check — the name and doc invite it. (Closely related to audit ref **L4**, the attenuation gap in the same method.)

## Recommended fix

Either (a) in `VerifyDelegationProofAsync`, after extracting the embedded parent (~lines 172-185), also check the parent's revocation; or (b) remove the leaf revocation check at line 126 and document that revocation is enforced only by `VerifyCapabilityChainAsync` / `VerifyInvocationAsync`. Tighten the `IVerificationService` doc so callers do not mistake the proof check for an authorization gate.

---
**How found:** adversarial multi-agent audit (correctness). Related: **L4** (attenuation not enforced in the same method).
**Audit ref:** M6

moisesja:
## Summary

Fixes #63 (M6). `VerifyCapabilityProofAsync` (the standalone single-proof "is this valid?" check) honoured revocation only for the **leaf** and its **immediate parent** — so a capability whose **root or an intermediate ancestor** had been revoked still passed, while `VerifyCapabilityChainAsync` (which checks every link) correctly rejected it. A consumer treating the single-proof method as a validity gate could accept a capability with a revoked ancestor.

A follow-up review of this branch found the first fix only closed the gap at **depth-1** (immediate parent), plus four lower-severity items. This PR addresses all of them.

## Fix

- **#1 (Medium) — full-ancestry revocation.** Revocation checking is centralized at the two entry points. The standalone path now sweeps **every** ancestor id carried in the delegation proof's `capabilityChain` (root + all intermediates + immediate parent, de-duplicated) via a new `IsAnyAncestorRevokedAsync`; the chain walk continues to revocation-check every resolved link. Both paths now reject a revoked ancestor at **any depth**. The now-redundant per-proof depth-1 check was removed from the shared single-delegation-proof verifier (now purely signature + parent authorization).
- **#2 (Low) — doc.** `RevokeCapabilityAsync` XML doc now states that authorization requires a fully valid chain: an expired capability, or one with a revoked ancestor (already inert), cannot be explicitly re-revoked.
- **#3 (Low) — nonce ordering.** The replay nonce is consumed **only after** the durable revocation write succeeds, so a throwing/failed store write no longer burns the request id against a legitimate retry.
- **#4 (Nit) — redundant builds.** Split `VerifyBuiltChainAsync`; `IsRevokerAuthorizedAsync` and `VerifyInvocationAsync` build the delegation chain once and reuse it.
- **#5 (Nit) — chain dedup.** `BuildCapabilityChain` de-duplicates ancestor ids (`[rootId, rootId, …]` → `[rootId, …]`).

## Verification

- New tests `VerifyCapabilityProof_WithRevokedRootAncestor_` and `_WithRevokedIntermediateAncestor_ShouldReturnFalse` (3-deep root revoked; 4-deep intermediate revoked) assert **both** `VerifyCapabilityProofAsync` and `VerifyCapabilityChainAsync` return `false`. Existing `_WithRevokedImmediateParent_` retained.
- Confirmed both new tests **fail** without the sweep (at `VerifyCapabilityProofAsync(leaf) == True`, the pre-fix bug) and **pass** with it.
- Full suite green: **379 Core + 6 AspNetCore**, 0 failed.
- Docs updated: `CHANGELOG.md` (the prior #63 entry overstated the depth-1 fix the same way the inline comment did) and `docs/SECURITY-FIXES-SUMMARY.md`.

> Doc-tightening on this method (clarifying it is not a full authorization gate) and attenuation enforcement are handled in the closely-related #69. Cumulative branch — diff includes #58–#62 until they merge.


moisesja (commented):
## Review of PR #83 — `fix(#63): honour ancestor revocation in VerifyCapabilityProofAsync`

CI: core build ✅ / ASP.NET build still in progress at review time.

The core fix is correct and well-motivated: `VerifySingleDelegationProofAsync` was the only path that skipped ancestor revocation, creating an inconsistency with `VerifyCapabilityChainAsync`. The new check (lines 209–214 in `VerificationService.cs`) closes that gap precisely and the regression test (`VerifyCapabilityProof_WithRevokedImmediateParent_ShouldReturnFalse`) verifies both paths now agree. Good work.

A few things worth discussing before merging:

---

### 1. Two breaking changes need explicit marking in CHANGELOG and semver version

**a) `InMemoryNonceStore` as default in convenience constructors (Issue #62)**

Changing the 2/3/4-arg constructors from `NullNonceStore` to `InMemoryNonceStore` is a **behaviour-breaking change** for any deployed consumer who relied on the previous (insecure) default. A stateless verifier running across multiple processes would silently start rejecting valid non-replayed invocations once the process restarts (the in-memory store is ephemeral). The CHANGELOG describes the issue and intent well, but the entry should be under `### Breaking Changes` (or similar), and this should trigger a **minor** semver bump at minimum (or patch if you consider it a security fix and document it as such). Is there a semver policy for this project?

**b) `RevokeCapabilityAsync(Capability, string)` added to `IVerificationService`**

Adding a new method to a public interface is source-and-binary-breaking for any existing external implementors. For now this is internal to the solution, but once this becomes a NuGet package any consumer who implements `IVerificationService` (e.g. a test double or custom store) will fail to compile. Worth noting explicitly in the CHANGELOG and considering a default interface method (C# 8+) if you want a softer landing: `Task<bool> RevokeCapabilityAsync(Capability capability, string revokerDid) => throw new NotImplementedException();`.

---

### 2. `IsRevokerAuthorizedAsync` silently swallows non-validation failures

```csharp
{
    return false;
}
```

This is correct for a malformed chain (can't authorize → deny), but exceptions from DID resolution failures, network errors (if a remote `IDidResolver` is in play), or crypto library errors will propagate as unhandled exceptions out of `RevokeCapabilityAsync(Capability, string)`. That might be the desired behaviour (fail-open on infrastructure errors so the host can decide), but it's an asymmetry worth documenting on the method. A comment or XML doc note along the lines of "exceptions from DID resolution or cryptography propagate; only `CapabilityValidationException` (malformed chain) yields `false`" would clarify intent.

---

### 3. Cumulative scope

The description calls this out, but the title says `fix(#63)` while the diff also includes changes for #59, #60, #61, and #62. If those have their own open PRs they'll conflict when they eventually land; if this is the canonical landing spot for all of them, the title and linked issues should reflect that. Minor — just making sure it's intentional.

---

### Minor positives worth noting
- The `EndToEndTests.cs` refactor (fresh invocation per call) is the right fix given replay-protection-on-by-default and avoids a false green from an in-process nonce collision.
- The SHOULD-04/SHOULD-05 compliance test renames correctly reframe spec intent: SHOULD measures at verify time, not create time; and `allowedAction` is application-defined vocabulary.
- Three-scenario coverage for the authorizing `RevokeCapabilityAsync` overload (authorized up-chain, unauthorized, self-revoke) is thorough.

Overall the fix is sound — the breaking-change documentation is the main thing I'd want addressed before this ships as a package release.

---

moisesja:
Thanks for the thorough review. Addressed in `32901a3`. One important thing surfaced while verifying:

### ⚠️ The branch did not compile (merge regression — now fixed)

The `fa87256` merge combined the two sides of this work incoherently. `d9410f3` (`fix(#60)`) had **removed** the bare-string `RevokeCapabilityAsync(string, string)` overload and migrated callers to the `Capability` overload; `0b10672` (the commit you reviewed) still had the bare-string overload and added the `#63` regression test against it. The merge kept the **new** interface (no bare-string overload) but left the `#63` test calling `RevokeCapabilityAsync(mid.Id, …)` and reverted CHANGELOG line 35 to text claiming the bare-string overload still exists. Net: `ZcapLd.Core.Tests` failed to compile (`cannot convert 'string' to 'Capability'`), so the "362 passed" no longer held on HEAD.

- Test migrated to `RevokeCapabilityAsync(mid, rootDid)` (rootDid is an up-chain controller of `mid`, so authorization succeeds). **Full suite: 366 passed.**
- CHANGELOG line 35 restored to the accurate "**BREAKING**: bare-string overload removed entirely" wording.

### 1. Breaking-change labeling — done

- **(a) Replay default (#62)** — now flagged `**BREAKING (behaviour)**` under Changed (cross-referencing the detailed Fixed entry), with the multi-node / process-restart caveat and the `NullNonceStore.Instance` opt-out spelled out.
- **(b) Revocation interface surface (#60)** — now flagged `**BREAKING (API)**`. Note the net change is *both* a removal (bare-string overload, per the merge fix above) and an addition (`Capability` overload), so I framed it as one surface change.

On **semver**: the Unreleased version is already `3.0.0` (a major bump), which covers all of these breaks per the project's stated SemVer policy — so no further version change is needed, only the explicit labeling you asked for.

On the **default interface method** suggestion: I deliberately kept `RevokeCapabilityAsync(Capability, string)` a regular abstract member rather than a DIM. For a security *authorization gate*, a `=> throw`/`=> false` default trades a compile error (which forces an implementor to consciously wire up authorization) for a runtime landmine or a silent always-deny. In an already-major release the source break is in-policy, so the safer landing here is the compile error. Documented the rationale inline in the CHANGELOG — happy to switch to a DIM if you'd prefer the softer landing.

### 2. `IsRevokerAuthorizedAsync` swallowing failures

The premise that infra exceptions propagate doesn't actually hold on current HEAD: `IsRevokerAuthorizedAsync` calls `VerifyCapabilityChainAsync` **first** (outside the try), and that method has a blanket `catch { return false; }`. The only code inside the try is `BuildCapabilityChainAsync`, which does no DID resolution or crypto and throws exactly `CapabilityValidationException`. So DID-resolution / network / crypto errors all resolve to `false` (nothing recorded) — the method is **fail-closed end to end**. I documented this on the interface (`<remarks>`) and added a comment at the `catch` site explaining why only `CapabilityValidationException` is caught.

### 3. Cumulative scope

Confirmed intentional — this is the cumulative landing branch; the diff carries #58–#62 until they merge. Title stays `fix(#63)` as the headline change.

moisesja:
Addressed all five review findings in 7f2c7c6:

- **#1 (Medium)** — the depth-1 gap is closed: the standalone path now sweeps every ancestor id in `proof.capabilityChain` (root + intermediates + immediate parent), so a revoked root/grandparent fails `VerifyCapabilityProofAsync` too. New `WithRevokedRootAncestor_` / `WithRevokedIntermediateAncestor_` tests (proven to fail without the sweep).
- **#2** doc clarified (authorization needs full chain validity; already-inert capability can't be re-revoked).
- **#3** replay nonce consumed only after the durable write.
- **#4** chain built once and reused (`VerifyBuiltChainAsync`).
- **#5** ancestor ids de-duplicated in `BuildCapabilityChain`.

Full suite green (379 + 6). The overstated comment/title and the prior CHANGELOG entry (which had the same "ancestor" overstatement) are corrected.

Issue #70, opened 2026-06-02
[L5] Add a structured VerificationResult failure-reason channel to verify methods


moisesja:
## Summary

Every `IVerificationService` verify method returns a bare `bool`. All internally-distinguished failure modes — revoked, expired, bad signature, broken chain, unauthorized controller, caveat failure, replay, chain-too-long — collapse to `false` with no reason code, out-parameter, or `Try`-style channel. Combined with the catch-all blocks (audit ref **M7**), infrastructure faults are also indistinguishable from denial.

## Location

[`src/ZcapLd.Core/Services/IVerificationService.cs:15,23,34,41`](src/ZcapLd.Core/Services/IVerificationService.cs#L15)

```csharp
Task<bool> VerifyCapabilityProofAsync(Capability capability);
Task<bool> VerifyInvocationAsync(Invocation invocation, Capability capability);
Task<bool> VerifyInvocationAsync(Invocation invocation, Capability capability, Dictionary<string, object>? contextProperties);
Task<bool> VerifyCapabilityChainAsync(Capability capability);
```

## Impact

**Low** (observability/ergonomics; fail-closed is safe). Consumers that want to surface *why* a capability failed must re-derive the reason themselves.

## Downstream note

`net-wallet-sdk` had to reconstruct the chain capability ids itself to recover a granular `Revoked` vs `ChainInvalid` reason, precisely because `VerifyCapabilityChainAsync` returns only `bool`. A structured result would remove that workaround.

## Recommended fix

Add a `VerificationResult` (outcome enum + optional message) sibling overload, keeping the `bool` methods as thin wrappers for compatibility; at minimum log the reason before returning `false`. (Audit ref **M7** is the catch-block half of this; fix together.)

---
**Affected version:** `ZcapLd.Core` 2.1.1 (source verified at tag `core-v2.1.1`)
**How found:** adversarial multi-agent audit (API-contract). Related: **M7** (catch-all masks errors).
**Audit ref:** L5

moisesja:
Related: #64 (M7) — the catch-block half of this; a structured result plus logging addresses both.

PR #84, opened 2026-06-02
fix(#58–#64): observability, replay protection, revocation authorization, and spec correctness

### LINKED ISSUE #64: [M7] Verify methods swallow all exceptions, masking config/transient errors as invalid
## Summary

All three public verify methods wrap their body in a bare `catch { return false; }`. Fail-closed is the right *security* default, but discarding the exception means transient and configuration faults — a missing canonicalizer registration, an unsupported proof type, a DID-resolution network error — are reported identically to a tampered capability. With no logging anywhere in `ZcapLd.Core`, a misconfiguration silently denies **all** valid capabilities and nothing tells the operator why.

## Location

[`src/ZcapLd.Core/Services/VerificationService.cs:146-149`](src/ZcapLd.Core/Services/VerificationService.cs#L146-L149), [`:315-318`](src/ZcapLd.Core/Services/VerificationService.cs#L315-L318), [`:384-388`](src/ZcapLd.Core/Services/VerificationService.cs#L384-L388)

```csharp
catch
{
    // For other unexpected exceptions, return false
    return false;
}
```

Note the inconsistency: the sibling `ResolvePublicKeyAsync` (line 395) deliberately *propagates* exceptions, while `ResolveCanonicalizer` (line 649) throws `CryptographicException` that these catch blocks then swallow into a plain `false`.

## Reproduction

```csharp
// Construct a verifier whose suite provider has no canonicalizer registered for the proof's suite,
// or hand it a proof whose Type has no registered suite:
bool ok = await verification.VerifyCapabilityChainAsync(perfectlyValidCapability);
// → false, with no exception, no log — indistinguishable from "capability is invalid"
```

The same masking happens for a transient `IDidResolver` failure (e.g. a `did:web` HTTP timeout): the capability is reported invalid rather than "could not verify, retry."

## Impact

**Low–medium.** Not a security break (nothing fails *open*) and not a correctness break, but a real operability/diagnostics defect: callers cannot retry on infrastructure failure, cannot alert on misconfiguration, and a forgotten suite/canonicalizer registration silently denies everything.

## Recommended fix

Keep fail-closed, but stop discarding the cause:
- `catch (Exception ex)` and log via an injected `ILogger`.
- Let configuration exceptions (no canonicalizer registered, unsupported proof type) propagate instead of masquerading as denial.
- Optionally introduce a tri-state result or a typed `VerificationException` for transient failures so callers can distinguish "invalid" from "couldn't check." (This is the operability half of the broader bare-`bool` return wart — see audit ref **L5**; worth fixing together.)

---
**How found:** adversarial multi-agent audit (correctness/observability). Related: **L5** (no failure-reason channel).
**Audit ref:** M7

moisesja:
## Summary

Fixes #64 (M7, observability). All three public verify methods wrapped their body in a bare `catch { return false; }`. Fail-closed is correct, but discarding the exception meant transient/configuration faults — a missing canonicalizer registration, an unsupported proof type, a DID-resolution network error — were reported identically to a tampered capability. With no logging anywhere in `ZcapLd.Core`, a misconfiguration silently denied **all** valid capabilities and nothing told the operator why.

## Fix

- `VerificationService` accepts an optional `ILogger<VerificationService>` (defaults to `NullLogger` — no output, no behavior change when unused).
- The three top-level `catch` blocks now `catch (Exception ex)` and `LogWarning(ex, …)` before returning false. **Fail-closed is preserved.**
- `ZcapLd.AspNetCore`'s `AddZcapServices` wires the container's `ILogger<VerificationService>` into the verifier.
- Adds a lightweight `Microsoft.Extensions.Logging.Abstractions` dependency to `ZcapLd.Core`.

A structured result / typed-exception channel (so callers can distinguish "invalid" from "couldn't check") is the broader bare-`bool` concern tracked as #70.

## Verification

New `VerificationServiceTests.VerifyCapabilityChain_WhenItFailsClosed_LogsTheCause`: a malformed delegated capability makes the chain verifier fail closed; a capturing `ILogger` confirms it returns `false` **and** logs a warning carrying the exception. Full suite green: 363 passed.

> Cumulative branch — diff includes #58–#63 until they merge.


moisesja (commented):
CI is green (363 tests) and the overall direction is solid — logging on fail-closed, replay protection on by default, and the authorizing revocation overload are all the right calls. Three things worth addressing before merge:

---

## 🔴 `IsRevokerAuthorizedAsync` catches too narrowly — transient errors propagate as unhandled exceptions

`IsRevokerAuthorizedAsync` only catches `CapabilityValidationException`:

```csharp
{
    return false;
}
```

`BuildCapabilityChainAsync` can also throw `InvalidOperationException`, `JsonException`, DID-resolution network errors, and more. Any of those will propagate through the public `RevokeCapabilityAsync(Capability, string)` API as an unhandled exception — callers expecting `false` on authorization failure will crash instead.

Every other security boundary in `VerificationService` applies the correct pattern:

```csharp
catch (Exception ex)
{
    // fail closed; log cause so a transient/config fault is distinguishable
    return false;
}
```

`IsRevokerAuthorizedAsync` should do the same. Logging the cause here (consistent with #64's intent) would also help operators distinguish a chain-walk failure from a genuine "not authorized."

---

## 🟡 PR title covers only #64 but the diff spans #58–#63 as well

The title says `fix(#64)` but the CHANGELOG and commit carry fixes for seven issues. The PR body is transparent about the cumulative branch, but `git log --oneline` will attribute all of it to `fix(#64)`. Consider something like:

> `fix(#58–#64): observability, replay protection, revocation authorization, and spec correctness`

---

## 🟡 Convenience-constructor change is a behavioral breaking change worth calling out explicitly

Switching the 2/3/4-arg constructors from `NullNonceStore.Instance` to `new InMemoryNonceStore()` is the right security default. Worth adding a `### Breaking Changes` entry to the CHANGELOG so upgraders know: any caller that passed the same `Invocation` object twice (tests, retries, logging middleware) will now see the second call fail. The README comment helps, but a dedicated CHANGELOG section would be more visible to consumers of the NuGet package.

---

moisesja:
Thanks — all three addressed in 69e955f.

While wiring this up I also found the branch wasn't compiling: the #63 merge (`7f2c7c6`) refactored `VerifyCapabilityChainAsync` into `VerifyCapabilityChainAsync` + `VerifyBuiltChainAsync` and mechanically moved #64's logging block (which references `capability.Id`) into the helper, where only the built `chain` is in scope — `error CS0103`. That also left the public method a bare non-logging `catch`, so #64's own `VerifyCapabilityChain_WhenItFailsClosed_LogsTheCause` test couldn't pass. Fixed both.

**🔴 `IsRevokerAuthorizedAsync` catches too narrowly** — The whole method body is now wrapped in `catch (Exception ex)`, so *any* fault from the chain walk (unbuildable chain, `JsonException`, DID-resolution network error) fails closed as "not authorized" instead of propagating as an unhandled exception through the public `RevokeCapabilityAsync` API. It logs the cause, consistent with #64. Added the same fail-closed logging to the `RevokeCapabilityAsync` top-level catch. New regression test `RevokeCapability_WhenAuthorizationChainWalkFaults_FailsClosedAndLogs` proves a chain-walk fault returns `false` (no crash) and logs a warning carrying the exception.

**🟡 PR title** — Retitled to `fix(#58–#64): observability, replay protection, revocation authorization, and spec correctness`.

**🟡 Convenience-constructor breaking change** — Added a dedicated `### Breaking Changes` section at the top of `[3.0.0]` (the four `BREAKING (...)` bullets moved out of `### Changed`), led by the nonce-store change with the concrete impact you called out: any caller re-submitting the **same `Invocation` instance** (same `Id`) — retries, idempotency layers, logging middleware, tests — now sees the *second* call return `false`; mint a fresh `Invocation.Id` per request.

Full suite green: 381 Core + 6 AspNetCore.

moisesja:
Addressed all four in 2e33ff5.

**🟠 Finding 1 (medium) — Warning-level log amplification on attacker input.** Fixed with a single `LogFailedClosed(ex, …)` helper that all six fail-closed catches now route through. Severity is chosen by exception type:
- **Debug** for expected, attacker-drivable validation/parse failures — `CapabilityValidationException` (covers unbuildable chain, unsupported proof type, **and** unresolvable/malformed `verificationMethod`, since `DidKeyResolver` wraps everything into it), plus `FormatException`/`JsonException` and the sibling domain exceptions.
- **Warning** stays for genuinely unexpected faults an operator must act on (missing canonicalizer/crypto-suite registration → `CryptographicException`, transient DID-resolution/infra errors, any non-library exception).

The tricky part you implied: `MultibaseCodec.Decode` wraps a malformed `proofValue` into the **same** `ZcapLd.Core.Exceptions.CryptographicException` used for the missing-canonicalizer config fault — so they can't be split at the catch. Rather than demote that type (which would bury #64's flagship case), I retype the input-level case at its source: a new `DecodeProofValue` rethrows a bad `proofValue` as `CapabilityValidationException`, so all four cited vectors land on Debug while the config fault stays Warning. No public-API or observable-result change — every path still returns `false`; only the log level and that one internal exception type changed (confirmed no test asserts an exception *type* from a verify path).

**🟡 Finding 2 (low) — CS0419.** Qualified to `<see cref="VerifyInvocationAsync(Invocation, Capability)"/>`. Clean `--no-incremental` build is CS0419-free; the remaining 7 warnings are the pre-existing `examples/ZcapLd.Examples` nullable ones.

**🟡 Finding 3a (low) — overstated comment.** You're right — `RevokeCapabilityAsync`'s outer catch already fails closed, so no caller was ever at crash risk. Reworded to the accurate rationale: the inner catch keeps the authorization decision local and attributes the cause to that step. CHANGELOG framing softened to match.

**🟡 Finding 3b (low) — test didn't isolate the inner catch.** The revocation test now asserts the inner catch's specific message (`"IsRevokerAuthorizedAsync failed closed…"`), which the outer catch's (different) message wouldn't satisfy — so it'd fail if the inner catch were narrowed or removed. Added `VerifyCapabilityChain_WhenUnexpectedFaultOccurs_LogsAtWarning` (a `ThrowingRevocationService` raising `InvalidOperationException`) to guard the **Warning** direction of the seam too. The two existing #64 tests now assert Debug + the attributing method name.

**ℹ️ Finding 4 (nit) — cumulative-branch non-compiling state.** Acknowledged; noted in `tasks/lessons.md`. Agree on rebasing #64 to a minimal delta once #58–#63 land so it can be evaluated in isolation.

Full suite green: 382 Core + 6 AspNetCore.

moisesja:
Both residual nits addressed in e099e33.

**Empty/null `proofValue` → Warning — fixed.** Confirmed the cause: `MultibaseCodec.Decode` throws `ArgumentException` from its null/empty guard (line 43) **before** its own `try`, so `DecodeProofValue`'s `catch (CryptographicException)` never saw it — it reached `LogFailedClosed` as `ArgumentException` (not in the expected set) → Warning. Broadened the catch:

```csharp
catch (Exception ex) when (ex is CryptographicException or ArgumentException)
```

Precise by construction: within that `Decode(proof.ProofValue)` call the *only* source of `ArgumentException` is the null/empty guard, so an empty `proofValue` now retypes to `CapabilityValidationException` → Debug, matching the doc. New `VerifyInvocation_WithEmptyProofValue_LogsAtDebugNotWarning` asserts a Debug entry with the retyped inner `ArgumentException` **and no Warning entry** — it would have failed before the fix.

**Resolver-dependent classification — doc note, as you suggested.** Kept the conservative Warning default for an unresolvable `verificationMethod` (genuinely ambiguous — attacker noise vs. a transient resolver outage the operator must see). Added a `<para>` to `LogFailedClosed` stating that type-based classification is best-effort for third-party `IDidResolver`s; the bundled `DidKeyResolver` already maps resolution failures to `CapabilityValidationException` (Debug). I deliberately did **not** add `ArgumentException` to the *global* expected set — it's a legitimate config/programmer-error signal elsewhere (the null/empty-DID guards in `ResolvePublicKeyAsync`/`DidKeyResolver`) that should stay at Warning; only the localized `proofValue` decode case is retyped.

Full suite green: 383 Core + 6 AspNetCore; no new build warnings.

Issue #50, opened 2026-05-02
Generate and verify spec-exact capabilityChain shape


moisesja:
## Summary

The live W3C CCG ZCAP-LD v0.3 draft defines an exact, minimal `proof.capabilityChain` shape. The current `CapabilityService.BuildCapabilityChain` embeds the root capability for first-level delegations and can duplicate root/ancestor IDs for deeper delegations. Because `capabilityChain` is part of the signed proof payload, this non-spec shape can break strict cross-implementation verification.

Spec source of truth: https://w3c-ccg.github.io/zcap-spec/

## Spec Requirement

In the **Delegated Capability** section, the draft says:

> A capability delegation chain MUST be an array that includes the root zcap using its ID (i.e., by reference only, not embedded) and every other delegated zcap in its ancestry must be referenced by ID except for the parent delegated zcap, which MUST be fully embedded.

Immediately after that, it says:

> The capability delegation chain is ordered; the first entry MUST be the root zcap's ID and any other entries must be in the order of delegation from least recent to most recent.

In the fetched spec snapshot used during review, these appear around lines 208-209.

### First-level delegation

When delegating directly from a root zcap, the chain should contain the root zcap by ID only:

```json
"capabilityChain": [
  "urn:zcap:root:https%3A%2F%2Fexample.com%2Ffoo"
]
```

The root zcap must not be embedded.

### Second-level delegation

When delegating from a first-level delegated zcap, the root remains ID-only and the immediate parent delegated zcap is embedded:

```json
"capabilityChain": [
  "urn:zcap:root:https%3A%2F%2Fexample.com%2Ffoo",
  {
    "@context": [
      "https://w3id.org/zcap/v1",
      "https://w3id.org/security/suites/ed25519-2020/v1"
    ],
    "id": "urn:uuid:first-delegation",
    "parentCapability": "urn:zcap:root:https%3A%2F%2Fexample.com%2Ffoo",
    "controller": "did:key:...",
    "invocationTarget": "https://example.com/foo",
    "expires": "2026-06-01T00:00:00Z",
    "proof": {
      "...": "..."
    }
  }
]
```

### Third-level delegation

Intermediate delegated ancestors are referenced by ID, and only the immediate parent is embedded:

```json
"capabilityChain": [
  "urn:zcap:root:https%3A%2F%2Fexample.com%2Ffoo",
  "urn:uuid:first-delegation",
  {
    "id": "urn:uuid:second-delegation",
[...]

PR #85, opened 2026-06-02
test+docs(#65): pin array-controller behavior; document DID-doc resolution limit

### LINKED ISSUE #65: [M8] Controller modeled as single string; IsControllerAuthorized is string-equality only
## Summary

`Capability.Controller` is modeled as a single non-nullable `string`, and `IsControllerAuthorized` is a string-equality check. Two consequences: (1) a spec-permitted **array** `controller` cannot be represented and **throws `JsonException` on deserialize**; (2) authorization is sound only for `did:key` (where the DID is derived from the key) — for DID methods where the controller DID differs from the key's containing DID, or where a controller authorizes multiple keys, it never resolves the controller's DID document to confirm the verification method is actually controlled by it.

Not a forgery risk (signatures are verified independently), but a real conformance gap that silently rejects conformant capabilities.

## Location

- [`src/ZcapLd.Core/Models/Capability.cs:27`](src/ZcapLd.Core/Models/Capability.cs#L27) — `Controller` is `string`
- [`src/ZcapLd.Core/Services/VerificationService.cs:584-588`](src/ZcapLd.Core/Services/VerificationService.cs#L584-L588) — `IsControllerAuthorized`

```csharp
private bool IsControllerAuthorized(string verificationMethod, Capability capability)
{
    var did = verificationMethod.Split('#')[0];
    return did == capability.Controller || verificationMethod == capability.Controller;
}
```

The repo's own checklist (`ZCAP-LD-SPECIFICATION-REQUIREMENTS.md:794`, "Support controller as string or array") is unchecked.

## Reproduction

```csharp
// (1) Array controller — round-trips in conformant ZCAP-LD elsewhere, throws here:
var json = """{ "@context":"https://w3id.org/zcap/v1","id":"urn:zcap:root:...",
               "controller":["did:key:zA","did:key:zB"],"invocationTarget":"https://x/y" }""";
var cap = JsonSerializer.Deserialize<Capability>(json, ZcapJsonOptions.Default);  // throws JsonException

// (2) Controller DID ≠ key-containing DID (e.g. did:web controller, key under a different DID):
//     IsControllerAuthorized compares strings only and never resolves the controller doc,
//     so a verification method legitimately authorized by the controller is rejected.
```

## Impact

**Low–medium (leaning medium for spec coverage).** Silently rejects array-controller capabilities and non-`did:key` controller arrangements. No forgery is enabled because the signature itself is verified.

## Recommended fix

- Model `controller` as string-or-array (custom `JsonConverter`, verbatim round-trip for JCS); apply OR semantics across entries.
- For non-`did:key` correctness, resolve the controller's DID document and confirm the verification method appears in the relevant verification relationship (`capabilityDelegation` / `capabilityInvocation`); this requires extending `IDidResolver`.
- Drop or tighten the loose `verificationMethod == Controller` fallback.
- Add array-controller and `did:web` (controller DID ≠ VM-containing DID) test vectors.

---
**How found:** adversarial multi-agent audit (spec-compliance/security). Array-controller deserialize throw empirically confirmed.
**Audit ref:** M8

moisesja:
## Summary

Addresses #65 (M8). The **empirically-confirmed** defect — a spec-permitted array `controller` threw `JsonException` on deserialize under the old single-`string` model, and `IsControllerAuthorized` was string-equality only — is already **resolved** by the `ControllerSet` work merged for #47: array controllers round-trip (single→bare string, array→array, shape-preserving for JCS), and authorization succeeds for any controller in the set (`ContainsVerificationMethod`).

This PR closes the loop:

- **Regression test** `RootInvocation_OverDeserializedArrayController_AuthorizesAnyController` — deserializes an array-`controller` capability from JSON (no throw) and authorizes an invocation signed by the **second** controller, end-to-end over the wire body.
- **Documents the remaining limitation** on `IsControllerAuthorized`: authorization is at DID-string granularity and does **not** resolve the controller's DID *document* to confirm the verification relationship for non-`did:key` controllers (controller DID ≠ verification-method DID). No forgery is enabled (the signature is always verified independently).

## Deferred (architectural)

The DID-document verification-relationship resolution explicitly **requires extending `IDidResolver`** — the same architectural class deferred for #50/#51. It's documented in code and tracked as the remaining part of #65 rather than bundled into this bounded change.

## Verification

Full suite green: 364 passed.

> Cumulative branch — diff includes #58–#64 until they merge.


moisesja (commented):
## Review — PR #85

Overall this is a well-reasoned, well-tested PR. Each behavioral change is backed by a pinning test and a clear spec citation. A few items worth addressing before merge:

---

### 1. `IsRevokerAuthorizedAsync` catch scope is narrower than its documented contract

**`VerificationService.cs` — `IsRevokerAuthorizedAsync`**

```csharp
{
    return false;
}
```

The XML doc on `RevokeCapabilityAsync(Capability, string)` says _"Returns false (recording nothing) when the revoker is not authorized **or the chain cannot be verified**."_ That promise covers any fault during chain resolution (DID resolution errors, transient network failures, missing canonicalizer, etc.), but the catch only covers `CapabilityValidationException`. Any other exception propagates through `RevokeCapabilityAsync(Capability, string)` as an unhandled fault — inconsistent with both the documented contract and the defensive pattern used everywhere else in the service.

Suggest widening to `catch (Exception)` (or `catch (Exception ex)` + `_logger.LogWarning`) to match the documented behavior, the same pattern used in the three public `Verify*` methods after issue #64.

---

### 2. Behavior-breaking default change (`NullNonceStore` → `InMemoryNonceStore`) is not marked `BREAKING` in CHANGELOG

The convenience-constructor replay-protection fix (#62) is a **behavioral breaking change** for any consumer using the 2–4-arg constructors today (previously zero replay protection; now replay protection on, and a process-local nonce store is allocated on every construction). The CHANGELOG entry is accurate but lives under `### Fixed` with no `BREAKING` callout. Consider adding a `⚠️ BREAKING` note or a `### Breaking Changes` subsection so consumers see it immediately on upgrade.

---

### 3. `CapturingLogger<T>` lives in the wrong file

The `CapturingLogger<T>` test helper class was added at the bottom of `VerificationServiceTests.cs`. It's already generic (`ILogger<T>`) and will likely be needed by future tests for other services. Moving it to `tests/ZcapLd.Core.Tests/Helpers/` alongside `InMemoryDidProvider` would keep it discoverable and consistent with the project structure.

---

### 4. Cumulative branch — review surface

The PR description notes _"Cumulative branch — diff includes #58–#64 until they merge."_ That's a reasonable approach to avoid conflicts, but it means 14 files and ~400 additions are being reviewed at once, most of which belong to prior issues. If any of #58–#64 are blocked, they block this one. No action needed, just flagging so the merge sequencing is deliberate.

---

### 5. CI still in progress

Both `build-test-pack-core` and `build-test-pack-aspnet` were in-progress at review time. Items 1–3 above are independent of CI outcome.

---

**Summary**: Items 1 and 2 are the substantive concerns. Item 1 is a correctness gap between the documented contract and the implementation. Item 2 is a changelog hygiene issue for a behavioral breaking change. Item 3 is a minor cleanup. Everything else looks solid.

---

Issue #72, opened 2026-06-02
[L7] CreateRootCapabilityAsync silently drops expires/allowedActions/caveats


moisesja:
## Summary

`CreateRootCapabilityAsync` accepts `expires`, `allowedActions`, and `caveats` parameters but silently drops all three — root capabilities are intentionally built without them (per W3C ZCAP-LD and the Issue #37 wire-shape). This is the create-path twin of the interface/impl signature wart (audit ref **L1**): the signature advertises inputs the method ignores. A caller passing `expires` for a root reasonably expects it to take effect; it does not.

## Location

[`src/ZcapLd.Core/Services/CapabilityService.cs:29-65`](src/ZcapLd.Core/Services/CapabilityService.cs#L29-L65) — `expires` / `allowedActions` / `caveats` accepted, never assigned (comment at line 61: fields "intentionally unset").

## Reproduction

```csharp
var root = await capabilities.CreateRootCapabilityAsync(
    aliceDid, "https://example.com/doc",
    allowedActions: new[] { "read" },
    expires: DateTime.UtcNow.AddDays(1),
    caveats: new Caveat[] { new ExpirationCaveat { Expires = DateTime.UtcNow.AddHours(1) } });

root.Expires.Should().BeNull();         // dropped
root.AllowedAction.Should().BeNull();   // dropped
root.Caveat.Should().BeNull();          // dropped
```

Note this is *correct* behavior for roots (a root carrying `expires` is actively rejected by `ValidateCapabilityAsync`); the problem is the misleading signature, not the drop.

## Impact

**Low.** Doc/impl mismatch; no security gap (roots must not carry `expires` — that is by design). Do **not** "fix" by honoring `expires` on roots.

## Recommended fix

Remove the silently-dropped `expires` / `allowedActions` / `caveats` parameters from the root overload (or document them as ignored / throw if supplied). Pairs with **L1** (same method's interface signature).

---
**Affected version:** `ZcapLd.Core` 2.1.1 (source verified at tag `core-v2.1.1`)
**How found:** adversarial multi-agent audit (API-contract). Companion: **L1**.
**Audit ref:** L7

moisesja:
Companion: #66 (L1) — the interface/impl signature mismatch on the same `CreateRootCapabilityAsync` overload.

Issue #37, opened 2026-04-29
Root capability wire format emits empty/null optional fields, breaking parsers that distinguish absent from empty


moisesja:
## Summary

Sibling to [#34](https://github.com/moisesja/zcap-dotnet/issues/34) — independent bug class, same blast radius (cross-language interop with [`zcap-py`](https://github.com/moisesja/zcap-py)).

The `Capability` model defaults its optional collection fields to `Array.Empty<>()` and `CreateRootCapabilityAsync` doesn't override that. Result: every root capability emitted on the wire carries

```json
{
  "allowedAction": [],
  "caveat": [],
  "parentCapability": null,
  "expires": null,
  "proof": null
}
```

W3C ZCAP-LD says these fields are **optional** on root capabilities. Strict-but-spec-conformant parsers (zcap-py, others) reject empty arrays / null for fields whose grammar is "must be a non-empty list of strings" / "must be a non-empty string" when present. They do not reject the field's absence. zcap-dotnet's emit-as-empty-or-null behaviour breaks them.

The maintainer's own xmldoc on `Capability.AllowedAction` already flagged this:

> NOTE C-01: Per strict W3C interpretation, root capabilities might not need allowedAction/caveat, but we include them for practical use. **Consider using serialization options to omit when null/empty.**

This issue is the follow-through on that note.

## How it surfaced (post-#34)

After upgrading to `ZcapLd.Core 2.0.0` to land #34, the live cross-language run in [moisesja/privacy-clinical-ai-framework](https://github.com/moisesja/privacy-clinical-ai-framework) cleared the `SignatureVerificationError` and now hits zcap-py's parser before chain verification can run:

```
zcap_verification_failed action=query_tumor_staging
  reason='allowedAction' must be a non-empty list of strings
  cause=<none>
```

Parser code that fires: [`zcap_py/zcap/parser.py:_parse_optional_action_list`](https://github.com/moisesja/zcap-py/blob/main/src/zcap_py/zcap/parser.py):

```python
def _parse_optional_action_list(d):
    v = d.get(\"allowedAction\")
    if v is None:
        return None                                           # absent — fine
    if not isinstance(v, list) or not v or not all(isinstance(a, str) for a in v):
        raise ZcapParseError(
            \"'allowedAction' must be a non-empty list of strings\",  # ← []  fails here
            field=\"allowedAction\",
        )
    return v
```

[...]

PR #86, opened 2026-06-02
fix(#66): align CreateRootCapabilityAsync interface signature with impl

### LINKED ISSUE #66: [L1] CreateRootCapabilityAsync interface/impl signature mismatch (allowedActions)
## Summary

`ICapabilityService.CreateRootCapabilityAsync` declares `allowedActions` as a **required, non-nullable** `string[]`, but the implementation declares it `string[]? allowedActions = null` and never reads it for root capabilities (roots carry no `allowedAction` by spec). The interface therefore forces every caller to pass an array the implementation silently discards. Runtime behavior is correct; this is an API-honesty / ergonomics wart.

## Location

- Interface: [`src/ZcapLd.Core/Services/ICapabilityService.cs:22`](src/ZcapLd.Core/Services/ICapabilityService.cs#L22) — `string[] allowedActions` (required)
- Impl: [`src/ZcapLd.Core/Services/CapabilityService.cs:32`](src/ZcapLd.Core/Services/CapabilityService.cs#L32) — `string[]? allowedActions = null`, with the comment at line 61 noting the field is "intentionally unset" on the root.

The root wire-shape (no `allowedAction`) is pinned by the Issue #37 cross-language test `CrossLanguageJcsInteropTests.cs:188-216`, so the discard is correct — only the interface signature is misleading.

## Reproduction

```csharp
ICapabilityService caps = new CapabilityService(signing);

// Caller is forced by the interface to supply an array that is then ignored:
var root = await caps.CreateRootCapabilityAsync(
    aliceDid, "https://example.com/doc", Array.Empty<string>());  // value discarded

root.AllowedAction.Should().BeNull();   // confirms the argument had no effect
```

A consumer binding the interface type cannot express "no enumerated actions" without passing a throwaway value (`Array.Empty<string>()` or `null!`).

## Impact

**Low.** Cosmetic/ergonomic; behavior is correct.

## Downstream note

`net-wallet-sdk` already works around this at `src/NetWallet/Capabilities/CapabilityService.cs:60-63`, passing `allowedActions: Array.Empty<string>()` with a comment "so we never bet on the library not dereferencing the parameter." Once fixed, that workaround can be dropped.

## Recommended fix

Make `ICapabilityService.cs:22` `string[]? allowedActions = null` (matching the impl), or remove the parameter from the root overload entirely (cleaner but source-breaking). Document that it is ignored for roots.

---
**How found:** surfaced directly during `net-wallet-sdk` Phase 2 integration; confirmed by the audit. Companion create-path issue: **L7**.
**Audit ref:** L1

moisesja:
## Summary

Fixes #66 (L1, API-honesty). `ICapabilityService.CreateRootCapabilityAsync` declared `allowedActions` as a **required** `string[]`, but the implementation declared it `string[]? allowedActions = null` and never reads it for roots (a root carries no `allowedAction` per W3C ZCAP-LD — the root wire-shape is pinned by the #37 cross-language test). A consumer bound to the interface had to pass a throwaway array to express "no enumerated actions".

## Fix

Make the interface signature match the implementation: `string[]? allowedActions = null`, documented as ignored for roots. **Non-breaking** — callers passing an array still compile, and callers can now omit it via the interface.

## Verification

New `CapabilityServiceTests.CreateRootCapability_ViaInterface_AllowedActionsOptional`: creates a root through the `ICapabilityService` interface without `allowedActions`; `AllowedAction` is null. Full suite green: 365 passed.

> Companion `expires`/`caveats` "ignored on roots" documentation is #72. Cumulative branch — diff includes #58–#65 until they merge.


moisesja (commented):
## Review — PR #86

Good set of targeted spec-compliance fixes. All behavioral changes have tests, the doc-string improvements are clear, and the inline comments explaining *why* code was removed (the old `read`/`write` allow-list, the `MaxExpirationMonths` throw, etc.) will help future readers. A few things worth flagging before merging:

---

### 1. Cumulative scope vs. stated title

The title says `fix(#66)` but the diff covers nine issues (#58–#66). The description notes this up front, which is helpful, but it means a reader scanning git log won't know where to look if a bisect lands here. If the dependent PRs (#58–#65) haven't merged yet, consider whether they should land first (or squash-merge this into labeled commits). If this is the intended single-PR batch, at minimum the title should reflect it (e.g. `fix(#58–#66): compliance batch`).

---

### 2. `IsRevokerAuthorizedAsync` catches only `CapabilityValidationException`

```csharp
{
    return false;
}
```

`BuildCapabilityChainAsync` can also throw `ArgumentNullException` or `InvalidOperationException` (from chain-assembly) and any `IDidResolver` implementation can throw on a network timeout. Those will propagate straight through `RevokeCapabilityAsync(Capability, string)`, giving callers an unexpected exception rather than a `false`. Since all three other verify paths already fail-closed with a logged warning, it would be consistent to broaden this to `catch (Exception)` (or at least `catch (Exception ex)` + a `_logger.LogWarning`) so an authorization check failure is never a surprise throw.

---

### 3. Issue #62 is a silent breaking change for in-process callers

Switching the convenience constructors from `NullNonceStore.Instance` to `new InMemoryNonceStore()` is the right security default, and the README update and `VerifyInvocation_TwoArgConstructor_RejectsReplayByDefault` test are good additions. The `EndToEndTests` usage-count test was also correctly patched to create a fresh invocation per attempt. Just worth calling out explicitly in the CHANGELOG's **Breaking Changes** section (or a migration note) so anyone upgrading doesn't silently start seeing replay-rejected results in their own tests or staging environment.

---

### 4. `CapturingLogger<T>` lives in a service test file

The `CapturingLogger<T>` helper is defined at the bottom of `VerificationServiceTests.cs`. This is fine for one test, but if other test classes (revocation tests, chain tests, etc.) end up needing it the duplication will be awkward. The `tests/Helpers/` directory already exists for exactly this kind of shared test infrastructure — worth moving it there now while the footprint is small.

---

### Minor / non-blocking

- The `mergeable_state` shows `unstable` — worth confirming CI is actually green before merging, especially since the suite reportedly grew from the previous count to 365.
- The `ZcapLd.AspNetCore` `AddZcapServices` now passes `ILogger<VerificationService>` to the full constructor. Double-check that `ILogger<VerificationService>` is always registered in the container at that point (it is when `AddLogging()` / the host builder is used, but library docs should mention it).

Overall this is a solid batch of fixes with good test coverage. Items 2 and 3 are the only ones I'd address before merging; the rest are suggestions.

---

moisesja:
## Review follow-up

Worked through the review against the **current** branch state. Since the review (2026-06-02), the dependent PRs #58–#65 merged to `main` and this branch rebased, so the PR diff has narrowed to exactly the **#66** fix (`ICapabilityService.cs`, its test, and the CHANGELOG). Status of each item:

1. **Cumulative scope vs. `fix(#66)` title** — Resolved by the rebase. `git diff main...HEAD` is now only the #66 change, so the title is accurate and a bisect landing here points at the right fix.

2. **`IsRevokerAuthorizedAsync` catching only `CapabilityValidationException`** — Already addressed (landed via the #65-review work, now on `main`): `VerificationService.IsRevokerAuthorizedAsync` now `catch (Exception ex)` and fails closed through `LogFailedClosed` (severity-aware). Not in this PR's diff anymore.

3. **#62 silent breaking change needs a CHANGELOG note** — Already present: CHANGELOG `### Breaking Changes` documents replay-protection-on-by-default (`NullNonceStore` → `InMemoryNonceStore`) with a concrete migration note.

4. **`CapturingLogger<T>` → `tests/Helpers/`** — Non-blocking, and the affected test files are now on `main` (outside this PR's diff). Left as a follow-up to avoid widening this narrowly-scoped PR; the helper is still defined once and works cross-file within the test assembly.

**Fixed in this push** (`4649973`): a bad merge had left the Issue #66 entry **duplicated verbatim** in the CHANGELOG `### Fixed` section — collapsed to a single line. This was the only defect within the current PR scope.

Verification: `dotnet build` clean; full suite green — **394** Core + **6** AspNetCore tests, 0 failed.

---
_Generated by Claude Code_

PR #87, opened 2026-06-02
fix(#67): select verification method by #fragment in DidKeyResolver

### LINKED ISSUE #67: [L2] DidKeyResolver ignores #fragment, always returns the first verification method
## Summary

`DidKeyResolver.ResolvePublicKeyAsync` strips the `#fragment` from the input and returns `VerificationMethod.FirstOrDefault()` regardless of which method the fragment names. For an Ed25519 `did:key`, net-did emits **two** verification methods (Ed25519 signing key at index 0, derived X25519 key-agreement key at index 1), so resolving the X25519 fragment returns the **Ed25519** key. It works today only by ordering coincidence; it is a latent contract violation of the documented fragment-addressed `IDidResolver`.

## Location

[`src/ZcapLd.Core/Services/DidKeyResolver.cs:48`](src/ZcapLd.Core/Services/DidKeyResolver.cs#L48) (fragment stripped), [`:60`](src/ZcapLd.Core/Services/DidKeyResolver.cs#L60) (`FirstOrDefault()`)

```csharp
var baseDid = didOrVerificationMethod.Split('#')[0];     // fragment dropped
...
var vm = result.DidDocument.VerificationMethod?.FirstOrDefault()   // ignores the named method
    ?? throw new CapabilityValidationException(...);
```

## Reproduction

```csharp
var resolver = new DidKeyResolver();
var ed = await didKeyMethod.ResolveAsync(ed25519Did);
var vms = ed.DidDocument!.VerificationMethod!;   // [0]=Ed25519, [1]=derived X25519

var k0 = await resolver.ResolvePublicKeyAsync(vms[0].Id);   // Ed25519 key (correct)
var k1 = await resolver.ResolvePublicKeyAsync(vms[1].Id);   // returns Ed25519 key AGAIN ← wrong
// Expected: k1 is the X25519 key named by vms[1].Id's fragment.
```

## Impact

**Low.** No exploit in pure-ZCAP usage: the crypto suite is selected by `Proof.Type` (not by `resolvedKey.KeyType`, which the verify path never reads), `IsControllerAuthorized` authorizes at DID granularity, and no in-tree path resolves a non-index-0 fragment. It bites the first consumer that key-agreement-resolves or that has a DID document listing multiple signing keys.

## Recommended fix

When the input contains `#`, select the `VerificationMethod` whose `Id` equals the full URI (then fall back to fragment match), and throw if none matches — rather than `FirstOrDefault()`.

## Downstream note

`net-wallet-sdk`'s own resolver adapter had the identical bug and has been fixed (fragment-aware selection: exact id → fragment → throw). The fix pattern transfers directly to `DidKeyResolver`.

---
**How found:** the same latent bug was caught in `net-wallet-sdk`'s resolver during PR review; the audit confirmed it is also present in zcap-dotnet's own `DidKeyResolver`.
**Audit ref:** L2

moisesja:
## Summary

Fixes #67 (L2). `DidKeyResolver.ResolvePublicKeyAsync` stripped the `#fragment` and returned `VerificationMethod.FirstOrDefault()` regardless of which method the fragment named. For an Ed25519 `did:key` — which resolves to **two** methods (the Ed25519 signing key at index 0 and a derived X25519 key-agreement key at index 1) — resolving the X25519 fragment returned the **Ed25519** key. It worked only by ordering coincidence and violated the documented fragment-addressed `IDidResolver` contract.

## Fix

When the input carries a `#fragment`, select the method whose `Id` equals the full URI (falling back to a fragment match), and throw `CapabilityValidationException` if none matches — rather than substituting the first. A bare DID (no fragment) still resolves to the primary method.

## Verification

- `ResolvePublicKeyAsync_X25519KeyAgreementFragment_ReturnsX25519KeyNotEd25519`: confirms the two-method shape under NetDid 1.3.0 and that the X25519 fragment now resolves to the X25519 key. Verified it **fails** before the fix, **passes** after.
- `ResolvePublicKeyAsync_UnknownFragment_Throws`: a bogus fragment throws instead of silently returning the first method.
- Full suite green: 367 passed.

> Cumulative branch — diff includes #58–#66 until they merge.


moisesja (commented):
CI green (both jobs), 367 tests pass. The core fix is clean and correct.

**#67 (primary fix):** `DidKeyResolver` now does exact-URI match → fragment fallback → throw, rather than always returning index 0. `FragmentOf()` helper is clear and the two new integration tests cover both the two-method shape and the bogus-fragment path. Good.

**A few things worth flagging for the release notes / upgrade guide:**

1. **Behavioral breaking change (#62):** The convenience constructors (`new VerificationService(resolver, caveatProcessor)`, etc.) now default to `InMemoryNonceStore` instead of `NullNonceStore`. This is the right security call, but any consumer that currently reuses the same `Invocation` object across multiple `VerifyInvocationAsync` calls will start failing on the second call. The CHANGELOG entry explains the "why" well, but doesn't flag it as a consumer-visible behavior change — worth a note in an upgrade guide.

2. **Interface additions (#60 and #66):** A new `RevokeCapabilityAsync(Capability, string)` overload on `IVerificationService` and `allowedActions` becoming optional on `ICapabilityService.CreateRootCapabilityAsync`. Both are additive and callers aren't broken, but anyone who has rolled their own implementation of either interface will need to add the new member. Worth calling out alongside point 1.

3. **`Microsoft.Extensions.Logging.Abstractions` in Core (#64):** Lightweight and justified — no concerns.

4. **`IsRevokerAuthorizedAsync` authorization granularity:** The DID-string-only match without resolving the controller document for non-`did:key` methods is an acknowledged limitation already tracked as #65. The `remarks` block on `IsControllerAuthorized` documents it clearly.

Everything else looks solid — SHOULD-04/SHOULD-05 spec alignment, ancestor-revocation check (#63), fail-closed logging (#64), and the end-to-end array-controller regression (#65) are all well-motivated and well-tested. LGTM once the upgrade-guide notes are in.

---

PR #89, opened 2026-06-02
fix(#69): enforce attenuation/expiry in VerifyCapabilityProofAsync; tighten doc

### LINKED ISSUE #69: [L4] VerifyCapabilityProofAsync does not enforce attenuation

moisesja:
## Summary

Fixes #69 (L4). `VerifyCapabilityProofAsync` verified a delegation's signature and parent-controller authorization but did **not** enforce attenuation (child ⊆ parent) or expiry. A re-signed child that **expands** authority beyond its parent was accepted by the single-proof method, while `VerifyCapabilityChainAsync` correctly rejected it — inviting consumers to misuse the proof check as an authorization gate.

## Fix

- The shared single-delegation-proof verifier now also rejects an **over-broad** child (`ValidateAttenuation` against the embedded parent) and an **expired** child — alongside the parent-revocation check added in #63. This makes `VerifyCapabilityProofAsync` a complete single-link soundness check.
- Tightened the `IVerificationService` / impl XML docs to state explicitly that this method checks **one delegation link** and is **not** a full authorization gate; callers wanting an authorization decision should use `VerifyCapabilityChainAsync` / `VerifyInvocationAsync`.

## Verification

New `VerifyCapabilityProof_OverbroadChild_ShouldReturnFalse`: a hand-signed child expanding `["read"]` → `["read","write"]` (only attenuation invalid; signature, authz, revocation, key-type all valid) is now rejected. Verified failing before the fix, passing after. Full suite green: 369 passed.

> Cumulative branch — diff includes #58–#68 until they merge.


moisesja (commented):
Overall this is a solid set of fixes — every change is motivated by a concrete spec violation or security gap, each is regression-guarded, and CI is green (369 tests pass). A few things worth discussing before merge:

---

**1. Issue #62 is a behavioral breaking change hiding in `### Fixed`**

Switching the convenience constructors from `NullNonceStore` to `new InMemoryNonceStore()` is the right security call, but it will break any consumer who:
- creates a `VerificationService` via the short constructors, and
- verifies the same `Invocation` object (same `id`) more than once (e.g. a test that calls `VerifyInvocationAsync` twice without round-tripping through JSON or changing the `id`).

The fix to `EndToEndTests.cs` (creating a fresh invocation per `InvokeOnceAsync()` call) is exactly the kind of silent breakage downstream consumers will hit. This belongs in a `### Breaking Changes` (or at minimum `### Changed`) section in CHANGELOG, not `### Fixed`, so consumers know to audit their test setups. Suggested wording:

> **Breaking:** `VerificationService` convenience constructors (2-, 3-, and 4-argument) now default to `InMemoryNonceStore` instead of `NullNonceStore`. Code that verifies the same `Invocation` instance twice (e.g. assertions or retry logic) will receive `false` on the second call. Use distinct `id` values per invocation, or pass `NullNonceStore.Instance` explicitly to opt out.

---

**2. SHOULD-04 enforcement gap until Issue #73 lands**

After this PR there is no enforcement of the 3-month expiration ceiling anywhere — it was removed from the create path and the verify-path implementation is deferred to #73. The CHANGELOG mentions this, but the compliance matrix now shows `SHOULD-04` as "delegation does not throw on long expirations at create time" — which is accurate but a reader scanning the matrix could mistake it for compliant when it's actually a deliberate in-flight gap. A note like *(ceiling enforcement deferred to #73)* next to the matrix row would self-document the intent.

---

**3. `IsRevokerAuthorizedAsync` silently drops non-`CapabilityValidationException` exceptions**

All three public verify methods now log unexpected exceptions before returning `false` (Issue #64). `IsRevokerAuthorizedAsync` only catches `CapabilityValidationException`; a transient DID-resolution failure or `HttpRequestException` escapes the catch and propagates up through `RevokeCapabilityAsync(Capability, string)` instead of returning `false`. This is inconsistent with the fail-closed/log-the-cause pattern established everywhere else. Consider broadening the catch or catching `Exception` and returning `false` there too (with a log warning).

---

**4. `Microsoft.Extensions.Logging.Abstractions` in `ZcapLd.Core`**

This is a lightweight and widely-accepted dependency, so not a blocker — but for a library that targets WASM (`wasi-experimental`) and advertises minimal dependencies, it's worth a line in the release notes / ARCHITECTURE.md noting the new transitive dependency for consumers who don't use `Microsoft.Extensions.*`.

---

**Minor / nits (no action required):**

- The ordering in `VerifySingleDelegationProofAsync` — attenuation and expiry are checked before the parent-authorization and signature checks. This is defensively fine (fail fast before crypto), but note that the attenuation is evaluated against the embedded parent whose own signature hasn't been verified by this path, so a tampered parent payload could theoretically be presented to influence the attenuation outcome. In practice the signature check below will catch it; just worth being explicit in a comment that this method is a *single-link* check and not a chain-of-trust verification.
- The new `RevokeCapabilityAsync(Capability, string)` return value (`false` = not authorized / cannot verify) conflates two distinct failure modes. A structured result type would let callers distinguish "unauthorized" from "chain unverifiable", but this is tracked as part of the broader #70 work — no change needed here.

None of these are blocking — the core fix (#69) is correct, the security improvements are valuable, and the test coverage is thorough. Resolving items 1–3 before merge would prevent downstream surprise.

---

PR #90, opened 2026-06-02
fix(#71): enforce proof.created freshness in VerifyInvocationAsync

### LINKED ISSUE #71: [L6] VerifyInvocationAsync has no Proof.Created freshness check
## Summary

`VerifyInvocationAsync` never validates `invocation.Proof.Created` freshness. Replay is bounded only by the nonce store's TTL: once the nonce entry evicts (default 5 minutes with `InMemoryNonceStore`), the identical captured invocation re-verifies. With `NullNonceStore` (the default convenience-constructor wiring — audit ref **M5**), replay is unbounded. The signed `Proof.Created` timestamp exists but is ignored.

## Location

[`src/ZcapLd.Core/Services/VerificationService.cs:218-319`](src/ZcapLd.Core/Services/VerificationService.cs#L218-L319) — the invocation verify path; nonce step at lines 308-311. No read of `invocation.Proof.Created` anywhere in the path.

## Reproduction

```csharp
// With the default InMemoryNonceStore (5-min window):
bool first = await verification.VerifyInvocationAsync(captured, capability);  // true
// ... wait out the nonce TTL (entry evicts) ...
bool later = await verification.VerifyInvocationAsync(captured, capability);  // true again ← stale replay
```

`captured.Proof.Created` may be hours old; it is never checked.

## Impact

**Low.** Not a spec MUST (ZCAP-LD models replay on the invocation `id` nonce), so this is a soundness wart rather than a violation. Becomes more serious in combination with **M5** (replay off by default).

## Recommended fix

Reject when `invocation.Proof.CreatedAt` is null, in the future beyond an allowed clock skew, or older than the replay window (`now - Created > _nonceWindow`). Anything evictable from the nonce store is then already too old to replay. Optionally feed the signed time into `InvocationContext.InvocationTime` for caveat evaluation.

---
**How found:** adversarial multi-agent audit (security). Compounds with **M5**.
**Audit ref:** L6

moisesja:
## Summary

Fixes #71 (L6). `VerifyInvocationAsync` never validated `invocation.Proof.Created` freshness, so replay was bounded only by the nonce store's TTL — once a nonce entry evicted (default 5 min), the identical captured invocation re-verified. With replay protection off (pre-#62) it was unbounded. The signed `Proof.Created` timestamp existed but was ignored.

## Fix

Reject an invocation whose `proof.created` is:
- **missing**,
- **future-dated** beyond a one-minute clock-skew allowance, or
- **older than the replay window** (`now − created > nonceWindow`).

Anything evictable from the nonce store is then already too stale to replay.

## Verification

Tests replicate signing over a controlled `created` timestamp (the public signing API always stamps "now"):
- stale (10 min old) → rejected; future (10 min ahead) → rejected; fresh ("now") → verifies (positive control proving the signature is valid, so the rejections are due to freshness).
- Confirmed the stale/future tests fail without the check.
- Full suite green: 372 passed.

> Cumulative branch — diff includes #58–#69 (and #71) until they merge. #70 (structured result) is deferred; its observability minimum was delivered by #64.


moisesja (commented):
Both CI checks pass and all 372 tests are green. The core freshness fix (#71) is solid — `nowUtc` is captured once, the null guard on `createdAt` is correct, and `_nonceWindow` is reused consistently instead of introducing a second constant. The test helper `SignInvocationWithCreatedAsync` cleanly isolates timestamp control without touching the public signing API. A few things worth flagging:

**Cumulative scope vs. PR title**
The title names #71 but the diff spans #58–#71 across 18 files. The description calls this out, but having 13 security fixes under a single PR title makes it harder to bisect regressions or map a future bug back to a specific change. Not a blocker here since the description is thorough, but worth tightening on future branches.

**`FreshnessClockSkew` is not configurable**
The 1-minute constant (`private static readonly TimeSpan FreshnessClockSkew`) is reasonable for most deployments, but environments with loose NTP sync (e.g. some IoT or air-gapped nodes) may need to tune it. Worth tracking as a follow-up — the constructor already accepts `nonceWindow`, so a companion `clockSkew` optional parameter would be consistent.

**Replay protection on by default is a behavioral breaking change (#62)**
Switching convenience constructors from `NullNonceStore.Instance` to `new InMemoryNonceStore()` is the right call security-wise, but any consumer who instantiated `new VerificationService(resolver, caveatProcessor)` and relied on the previously permissive behavior (e.g. in a long-running process where the store's TTL would accumulate) will silently start seeing replay rejections after upgrading. The CHANGELOG documents this well; just confirm it's called out clearly in any release notes or migration guide.

**ASP.NET revocation endpoint gap (#60)**
The description flags it: the `ZcapLd.AspNetCore` revocation endpoint still accepts a bare `revokerDid` without authorization. Since the new `RevokeCapabilityAsync(Capability, string)` overload exists and the old bare-string overload is now documented as "attribution only", consumers of the endpoint could unknowingly still land on the unauthenticated path. Please make sure this is tracked as an issue so it doesn't get lost in the cumulative backlog.

**`ValidateAttenuation` refactored to return `bool`** — confirmed in the working tree; the call sites in `VerifySingleDelegationProofAsync` and the chain walk both handle the `false` return correctly. No concerns there.

---

PR #91, opened 2026-06-02
docs(#72): document that CreateRootCapabilityAsync ignores expires/allowedActions/caveats

### LINKED ISSUE #72: [L7] CreateRootCapabilityAsync silently drops expires/allowedActions/caveats

moisesja:
## Summary

Fixes #72 (L7, api-design). `CreateRootCapabilityAsync` accepts `expires`, `allowedActions`, and `caveats` but silently drops all three. This is **correct** — a root carries none of these per W3C ZCAP-LD (a root with `expires` is actively rejected by `ValidateCapabilityAsync`, and the root wire-shape is pinned by the #37 cross-language test). The wart is the misleading signature, not the drop.

## Fix

Removing the parameters is source-breaking and throwing on `allowedActions` would break every root-creation call site, so per the issue's accepted option the parameters are **documented as ignored / not honored**. The interface `expires`/`caveats` param docs now say so explicitly (companion to #66, which aligned the `allowedActions` signature). **Behavior is unchanged** — and deliberately so.

## Verification

New `CapabilityServiceTests.CreateRootCapability_DropsExpiresAllowedActionsAndCaveats`: supplies all three to a root and asserts `AllowedAction`/`Expires`/`Caveat` are null and the root still validates — locking the drop so a future change does not mistakenly start honoring them (the issue explicitly warns against honoring `expires` on roots). Full suite green: 373 passed.

> Cumulative branch — diff includes #58–#71 until they merge.


moisesja (commented):
## Review — PR #91

The individual fixes are well-reasoned, thoroughly tested, and each one is correctly scoped to its issue. Some systemic concerns worth addressing before merge:

---

### 1. Title/scope mismatch (cosmetic, but important for git history)

The title says `docs(#72)` but the diff addresses **14 issues** (#58–#72), including behavioral security fixes. The description acknowledges this ("cumulative branch — diff includes #58–#71 until they merge") but the commit title will be misleading in `git log`. Consider either:
- Retitling to reflect the actual scope (e.g. `fix(#58-#72): security hardening and doc fixes`), or
- Noting in the description which issues are *new* to this PR vs. carried from already-merged work.

---

### 2. Breaking interface change — `IVerificationService.RevokeCapabilityAsync(Capability, string)`

Adding a new method to a public interface is source-breaking for any consumer that has their own `IVerificationService` implementation (mocks, adapters, decorators). This isn't covered under a "Breaking Changes" section in `CHANGELOG.md`.

**Suggestion**: either add a "Breaking Changes" section to the changelog entry, or consider using a default interface implementation (`Task<bool> RevokeCapabilityAsync(Capability capability, string revokerDid) => ...`) to keep it non-breaking.

---

### 3. Issue #61 leaves a gap — 3-month expiration is now completely unenforced

The create-time check was removed (correct — it was the wrong actor/time), and the description says the ceiling "moves to the verify path behind a policy flag (Issue #73)" — but Issue #73 is not addressed in this PR. The compliance test was also renamed from "Verifier SHOULD reject delegated expirations beyond 3 months" to "does not hard-throw at create time", which passes now but the verifier-side enforcement is absent.

This leaves the SHOULD-04 requirement with zero enforcement in any code path. That's acceptable as a transitional state if #73 is coming soon, but worth a note in the changelog or test comment.

---

### 4. Behavior change for replay protection (Issue #62) — arguably breaking

Changing the 2/3-arg `VerificationService` constructors from `NullNonceStore` → `new InMemoryNonceStore()` is the right default security posture, but it's a behavior change: any caller relying on the old "pass the same invocation multiple times without rejection" behavior silently breaks. The README comment addresses multi-node callers, but the `CHANGELOG.md` entry under Fixed doesn't flag this as a potentially observable behavior change.

Minor suggestion: move this to a **Changed** entry (not just Fixed) since existing correct behavior now becomes incorrect.

---

### Minor notes (non-blocking)

- `IsRevokerAuthorizedAsync` checks `Controller.ContainsVerificationMethod(revokerDid)` — this relies on the same DID-string granularity documented as a limitation of `IsControllerAuthorized`. The inline comment on `IsControllerAuthorized` correctly notes the caveat. Consistent, just worth confirming the limitation is acceptable for the revocation path too.
- `FreshnessClockSkew = TimeSpan.FromMinutes(1)` is a constant — consider exposing it (or at least documenting it) so consumers know what clock drift is tolerated.
- The `CapturingLogger` and `WrongKeyTypeResolver` test helpers live in the main test file. If they grow, moving them to `Helpers/` would be consistent with the project structure.

---

### Overall

The security fixes (#62 replay default, #63 ancestor revocation, #68 key-type binding, #60 authorization check, #71 freshness) are correct and well-tested. The main ask before merge is addressing the **interface breaking change** documentation and the **Issue #73 enforcement gap** — or at minimum making them explicit in the changelog so consumers aren't surprised.

---

PR #92, opened 2026-06-02
feat(#73): add opt-in verifier-side 3-month expiration ceiling

### LINKED ISSUE #73: [L8] Verifier-side 3-month expiration SHOULD is absent

moisesja:
## Summary

Fixes #73 (L8, spec-compliance) — companion to #61. The W3C ZCAP-LD verifier-side SHOULD ("a verifier SHOULD ensure an invoked delegated zcap does not have an expiration more than three months in the future") was **absent** on the verify path; the 3-month ceiling lived (wrongly) as a create-time hard throw, removed in #61. This adds it where the spec places it.

## Fix

- New `VerificationPolicy` (`EnforceMaxDelegationExpiration`, default **false**; `MaxDelegationExpirationMonths`, default **3**).
- `VerifyCapabilityChainAsync` enforces the ceiling **per delegated link, at verification time**, when the policy is enabled.
- Off by default because it is a SHOULD and can reject legitimately long-lived delegations a parent permitted.
- Supplied via the full `VerificationService` constructor; the ASP.NET DI factory resolves an optional registered `VerificationPolicy` (`services.AddSingleton(new VerificationPolicy { EnforceMaxDelegationExpiration = true })` before `AddZcapServices`).

## Verification

The previously-misnamed `SHOULD-04` compliance test now genuinely exercises the verifier: the same 6-month delegation **verifies with the policy off** and is **rejected with it on**. Matrix repointed to the verifier test (alongside #61's create-time no-throw test). Full suite green: 374 passed.

> Cumulative branch — diff includes #58–#72 until they merge.


moisesja (commented):
**Code review — looks good.** A dense cumulative diff, but each change is well-reasoned, spec-grounded, and regression-tested. A few observations below.

---

### What's good

**Security posture improvements (Issues #62, #68, #71)**
- Flipping convenience constructors from `NullNonceStore` to `InMemoryNonceStore` is the right secure-by-default call. The Quick-Start README path now gives replay protection out of the box.
- The `proof.created` freshness window bounds replay independently of nonce-store TTL — a genuine defense-in-depth gap is now closed.
- Explicit key-type/suite binding (`suite.KeyType != resolvedKey.KeyType`) self-documents the invariant and future-proofs against custom resolvers; the `WrongKeyTypeResolver` test helper makes the intent clear.

**Spec correctness (Issues #59, #61, #73)**
- Moving the 3-month ceiling from a create-time hard throw to a verifier-side opt-in `VerificationPolicy` is the correct reading of the ZCAP-LD SHOULD. The dual test (policy-off verifies, policy-on rejects) cleanly demonstrates both behaviors.
- Removing the closed `read`/`write` allow-list from `ValidateCapabilityAsync` is correct; the spec is explicit that `allowedAction` is application-defined vocabulary.

**`VerifyCapabilityProofAsync` semantics (Issues #63, #69)**
- The single-link check now honors parent revocation and attenuation — removing the incoherence with `VerifyCapabilityChainAsync` where the standalone method could accept what the chain check would reject.

**`DidKeyResolver` fragment fix (Issue #67)**
- The full-URI-then-fragment fallback is the right approach. The regression tests for the X25519 method and an unknown fragment are exactly what's needed here.

**Logging (Issue #64)**
- `ILogger<VerificationService>` defaulting to `NullLogger` keeps the Core library dependency-light (just the Abstractions package) while making misconfiguration/transient faults observable. The `CapturingLogger` test helper is clean.

---

### Minor observations

1. **`IsRevokerAuthorizedAsync` catches only `CapabilityValidationException`**: other unexpected exceptions (DID resolution network errors, etc.) bubble through `RevokeCapabilityAsync(Capability, string)` rather than returning `false`. This is arguably the right choice — a transient fault shouldn't silently be misread as "unauthorized" — but worth noting on the interface doc so callers know to expect it.

2. **`FreshnessClockSkew` (1 minute) is a private constant**: reasonable for now, but as this reaches multi-region production deployments, 1 minute can be tight. Consider exposing it on `VerificationPolicy` (or at minimum noting the value in the XML doc) so operators can widen it without a fork.

3. **Cumulative branch scope**: noted in the description (#58–#72 bundled). Large surface area, but each change is independently motivated and tested, which makes it workable. The compliance matrix update is a nice touch.

374 tests passing, all changes traceable to spec or security rationale, test coverage on every new code path.

---

PR #93, opened 2026-06-02
fix(#74): use ordinal StartsWith in invocation-target prefix check

### LINKED ISSUE #74: [L9] Culture-sensitive StartsWith in invocation-target prefix check
## Summary

The invocation-target prefix test uses a culture-sensitive `string.StartsWith` overload while the surrounding logic (and the `Substring` that follows it) is ordinal. With ignorable Unicode characters the two can disagree, and a corner case throws `ArgumentOutOfRangeException`. Both call sites catch and return `false` (fail-closed), and real targets are ASCII URIs, so impact is minimal — but it is a latent correctness/consistency bug.

## Location

[`src/ZcapLd.Core/Services/VerificationService.cs:556`](src/ZcapLd.Core/Services/VerificationService.cs#L556)

```csharp
// culture-sensitive overload, inconsistent with the ordinal Substring below it
if (invocationTarget.StartsWith(capabilityTarget))
{
    var suffix = invocationTarget.Substring(capabilityTarget.Length);   // ordinal index math
    ...
}
```

`StartsWith(string)` is culture-sensitive; it can return `true` for strings whose ordinal lengths differ (ignorable code points), after which `Substring(capabilityTarget.Length)` can be wrong or throw `ArgumentOutOfRangeException`.

## Reproduction

Construct `capabilityTarget` / `invocationTarget` differing only by a Unicode ignorable character such that the culture-sensitive `StartsWith` returns `true` but the ordinal lengths mismatch; the subsequent `Substring` then misbehaves. (Caught by the surrounding `try`/`catch` → `false`, so it manifests as a spurious rejection rather than a crash.)

## Impact

**Low.** Fail-closed and real targets are ASCII URIs. The matching algorithm itself is spec-correct; only the comparison overload is wrong.

## Recommended fix

```csharp
if (invocationTarget.StartsWith(capabilityTarget, StringComparison.Ordinal))
```

This also removes the `ArgumentOutOfRangeException` corner case.

---
**How found:** adversarial multi-agent audit (spec-compliance). Trivial fix; good first issue.
**Audit ref:** L9

moisesja:
## Summary

Fixes #74 (L9, good first issue). The invocation-target prefix test used the **culture-sensitive** `string.StartsWith(string)` overload while the following `Substring(capabilityTarget.Length)` is ordinal. With ignorable Unicode code points the two can disagree, and a corner case throws `ArgumentOutOfRangeException` (caught by the surrounding try/catch → spurious rejection).

## Fix

One line: `invocationTarget.StartsWith(capabilityTarget, StringComparison.Ordinal)` — consistent with the ordinal `Substring` index math, and free of the `ArgumentOutOfRangeException` corner case.

## Note on testing

Impact is minimal (both call sites fail closed; real targets are ASCII URIs), and the fix has **no observable change in the boolean result** — the value is correctness/consistency and removing the masked exception. The added test (`VerifyInvocation_TargetWithIgnorableUnicodeChar_IsNotTreatedAsPrefix`) is a behavior-lock: a target with an injected soft hyphen (U+00AD) is not treated as an ordinal prefix and is rejected.

Full suite green: 375 passed.

> Cumulative branch — diff includes #58–#73 until they merge.


moisesja (commented):
## Review — PR #93

I read the full diff (1,489 lines, 19 files). The code quality is high, each fix is well-reasoned, and coverage is thorough. A few things worth discussing before merge:

---

### 1. PR scope vs. title (informational, not a blocker)

The title says `fix(#74)` but the branch is a cumulative batch of **16 separate fixes** (#59–#74), as the description notes. That's totally workable, but it means every item below deserves the same scrutiny as the headline fix. The CHANGELOG and test names are explicit enough that reviewing issue by issue is feasible.

---

### 2. Breaking interface change — `IVerificationService` (needs decision)

`IVerificationService` gains a new method:

```csharp
Task<bool> RevokeCapabilityAsync(Capability capability, string revokerDid);
```

Adding a method to a public interface is a **binary and source breaking change** for any consumer who implements the interface (e.g., test doubles, wrappers, mock objects). Does this library have a stable API policy / SemVer commitment? If so, this should bump the major version or the interface should become an abstract base class / default-interface-method for .NET 10.

---

### 3. Behavioral breaking change — replay protection on by default (#62)

```csharp
// Before: NullNonceStore.Instance
// After:  new InMemoryNonceStore()
```

This is the right security fix, but it is a **silent behavioral change** for existing consumers of the convenience constructors. Any code that was (accidentally) relying on replays being allowed — or that calls `VerifyInvocationAsync` with the same invocation object twice in tests — will now fail. The README comment and constructor XML doc explain the change clearly, but the release notes / CHANGELOG should call this out under **Breaking Changes**, not just **Fixed**.

---

### 4. Freshness window couples to nonce window (#71)

```csharp
if (nowUtc - createdAt.Value > _nonceWindow)   // _nonceWindow defaults to 5 min
    return false;
```

The staleness bound for `proof.created` is `_nonceWindow`, the same value that controls nonce tracking. This means a consumer who extends the nonce window (e.g., to 1 hour for long-lived service accounts) also silently extends the freshness acceptance window to 1 hour. That is probably unintentional — `proof.created` staleness and nonce TTL are different concerns. Consider a separate `MaxProofAge` (or similar) field on `VerificationPolicy`, so the two can be tuned independently.

---

### 5. `IsRevokerAuthorizedAsync` swallows only `CapabilityValidationException`

```csharp
{
    return false;
}
```

A DID resolution failure (`HttpRequestException`, `TaskCanceledException`) or a null-reference in chain traversal would propagate uncaught and bubble out of `RevokeCapabilityAsync(Capability, string)` as an unhandled exception — unlike the three public verify methods, which all fail closed. Consider either catching `Exception` here (fail closed, consistent with the verifier) or letting it throw (which is also fine, but then the XML doc should say so). Right now the behavior is inconsistent.

---

### 6. Minor — step labels out of order in `VerifyInvocationAsync`

The freshness check is inserted as `// 2c.` but it appears **before** the existing `// 2a.` check in the code. This doesn't affect correctness but will confuse the next reader.

---

### 7. Minor — tests pass ignored `allowedActions` to `CreateRootCapabilityAsync`

Several new tests (e.g., `NormativeIntegrationComplianceTests.Should04_…`, `VerifyCapabilityProof_OverbroadChild_…`) call:

```csharp
CreateRootCapabilityAsync(rootDid, "...", new[] { "read" })
```

Since `allowedActions` is documented as ignored for roots, `null` (or simply omitting it) would be clearer and model the intended usage better. Minor, but test code is documentation.

---

### Summary

| Issue | Severity |
|-------|----------|
| Interface break (`IVerificationService`) | Needs decision |
| Behavioral change (replay on by default) needs CHANGELOG entry | Low |
| Freshness vs. nonce window coupling | Low–Medium |
| `IsRevokerAuthorizedAsync` partial exception catch | Low |
| Step label order | Nit |
| Ignored-arg clarity in tests | Nit |

The core fix (#74) is correct and the regression test is good. The larger security fixes (#62, #63, #67, #68, #69, #71) are all sound. Main ask before merging: a decision on the interface versioning story and the two minor behavioral concerns noted above.

---

PR #94, opened 2026-06-02
fix(#61): address PR #81 review — revoker auth semantics + SHOULD-04 assertion

moisesja:
## Summary

Follow-up to the **merged** PR #81. Its review raised two items to address before merge; #81 merged before these fixes landed, so they ship here as a small follow-up. No production behavior change beyond comments — **364 tests green**.

## Changes

**1. Tighten the SHOULD-04 compliance test**
The test only asserted `delegated.Expires` was non-null (i.e. "no exception thrown"). It now asserts the requested 6-month expiration round-trips intact, catching a silent truncation in `DelegateCapabilityAsync`:
```csharp
delegated.ExpiresAt.Should().NotBeNull();
delegated.ExpiresAt!.Value.Should().BeCloseTo(DateTime.UtcNow.AddMonths(6), TimeSpan.FromSeconds(5));
```
(The reviewer suggested asserting on `Expires`, but that's the verbatim on-wire `string?`; the parsed `DateTime?` view is `ExpiresAt`.)

**2. Document `IsRevokerAuthorizedAsync` authorization semantics**
The review flagged a possible false-negative for `did:web` controllers with key fragments. On inspection, the cited example (`controller=did:web:issuer.example`, `revoker=did:web:issuer.example#key-1`) actually **passes** — `ContainsVerificationMethod` splits on `#`, so the bare DID matches. The real gap is **cross-DID** authorization: a revoker key belonging to a *different* DID the controller's DID document would authorize is not matched, because no DID resolution happens.

- Added a `TODO` + clarifying comment in `VerificationService.IsRevokerAuthorizedAsync` describing exactly what the string match covers and what it doesn't.
- Added two `ControllerSetTests`: one proving the key-fragment case works, one pinning the cross-DID limitation as known behavior.

## Not addressed (with reasoning)

- **Warning log for long expirations** — optional/conditional in the review. `CapabilityService` has no logger and Core has zero logging infrastructure; adding a package dependency + constructor change to land a soft suggestion is scope creep against #81's deliberate decision to enforce the ceiling on the verify path in #73.
- **Cumulative branch note** — informational only.

### DIFF
--- a/src/ZcapLd.Core/Services/VerificationService.cs
+++ b/src/ZcapLd.Core/Services/VerificationService.cs
@@ -726,6 +726,14 @@ public class VerificationService : IVerificationService
         try
         {
             var chain = await BuildCapabilityChainAsync(capability);
+            // Authorization is a string-level controller match: revokerDid (a bare DID or a
+            // did#key-fragment) authorizes when it equals a chain controller, or its bare DID
+            // does. This covers did:key (the DID *is* the key) and a did:web controller whose
+            // bare DID matches the revoker's key fragment (did:web:issuer#key-1 → did:web:issuer).
+            // It does NOT resolve the controller's DID document, so a revoker key belonging to a
+            // *different* DID that the controller would authorize is not matched.
+            // TODO: support cross-DID key authorization by resolving the controller's DID
+            // document and checking its verificationMethod/capabilityDelegation relationships.
             return chain.Any(link =>
                 link.Controller is not null &&
                 link.Controller.ContainsVerificationMethod(revokerDid));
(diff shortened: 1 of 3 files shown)


moisesja (commented):
Overall this is a clean, well-scoped follow-up. The SHOULD-04 assertion is now actually meaningful, the two new `ControllerSetTests` clearly document behavior vs. known limitation, and the `VerificationService` comment captures a genuinely non-obvious authorization semantic. No concerns with the logic.

**One minor issue to address before merge:** several comments reference `(PR #81 review)`. Per the project's own CLAUDE.md: *"Don't reference the current task, fix, or callers… since those belong in the PR description and rot as the codebase evolves."* Three spots:

- `NormativeIntegrationComplianceTests.cs`: `// that a NotBeNull() check would miss (PR #81 review).`
- `ControllerSetTests.cs` (key-fragment test): `// (PR #81 review: IsRevokerAuthorizedAsync key-fragment edge case.)`
- `ControllerSetTests.cs` (cross-DID test): `// (see the TODO in VerificationService.IsRevokerAuthorizedAsync). (PR #81 review.)`

The PR description already carries that history perfectly. The comments would read more cleanly without the parenthetical PR citation. Everything else looks good — happy to approve once those are trimmed.

---

PR #96, opened 2026-06-05
fix: generate and verify spec-exact capabilityChain (#50)

### LINKED ISSUE #50: Generate and verify spec-exact capabilityChain shape

moisesja:
## Summary

The delegation proof's `capabilityChain` now follows the live W3C CCG ZCAP-LD draft **exactly**: the root is referenced **by id only** (never embedded), every ancestor by id, and **only the immediate parent** delegated zcap is fully embedded (last entry). Generation previously embedded the root at first level and emitted the immediate parent *both* as an id string and embedded; the verifier was hard-coupled to that non-spec shape and could not resolve a root referenced by id. Because `capabilityChain` is inside the signed proof, the old shape broke strict cross-implementation verification.

The issue was re-confirmed with a multi-agent evaluation (4 investigation dimensions + 3 adversarial skeptics, all high-confidence — one ran the live code and observed chains of length 2/3/4 instead of the spec's 1/2/3).

## What changed

| Level | `capabilityChain` (now) |
| --- | --- |
| First | `[rootId]` |
| Second | `[rootId, {parent}]` |
| Third | `[rootId, firstId, {parent}]` |

- **Generation** — `CapabilityService.BuildCapabilityChain` emits the minimal spec-exact shape (reuses the existing id extractor; drops the dedup + double-emit).
- **Root resolution** — new `IRootCapabilityResolver` (+ `InMemoryRootCapabilityResolver` dev impl) and explicit-root overloads on `VerifyCapabilityChainAsync` / `VerifyCapabilityProofAsync` / `VerifyInvocationAsync` / `RevokeCapabilityAsync`. Precedence: **explicit root → resolver → fail-closed**, with root-id and `urn:zcap:root:` target-encoding binding to prevent substitution. An `IDidResolver` that also implements `IRootCapabilityResolver` is auto-detected (mirrors the existing relationship-resolver pattern).
- **Strict verification** — both the chain-walk and standalone single-proof paths share one validator that rejects an embedded root, duplicate ids, a parent referenced both by id and embedded, and a wrong/missing embedded parent.
- **ASP.NET / examples** — `AddZcapRootCapabilityResolver<T>()` + DI wiring; console and revocation-demo examples register roots.
- **Docs** — README (new “Delegation Chains & Root Resolution” section + migration note), ARCHITECTURE, AGENTS.

## ⚠️ Breaking change (3.0.0)

The `capabilityChain` wire format no longer embeds the root, so capabilities signed by earlier versions must be **re-delegated**, and verifying a delegated capability now requires a configured `IRootCapabilityResolver` **or** an explicit root passed to the verify/revoke overloads.

## Test plan

- `dotnet build ZcapLd.sln` — clean (Debug + Release, 0 code warnings).
- `dotnet test ZcapLd.sln` — **417 passing, 0 failing** (411 Core + 6 AspNetCore), including new `CapabilityChainShapeTests` (L1/L2/L3 generation, accept via resolver + explicit root, fail-closed without a root, and 4 non-spec rejection cases) and the existing cross-language JCS/RDFC interop fixtures (already encoded `[rootId]` — they ratify the fix, no change needed).
- `dotnet run --project examples/ZcapLd.Examples` — all delegated verifies print `True (expected: True)`.

Closes #50


moisesja (commented):
## Review — fix: generate and verify spec-exact `capabilityChain` (#50)

This is a well-executed, high-impact correctness fix. The chain shapes are now spec-exact, the `IRootCapabilityResolver` design is clean, the fail-closed semantics are consistent, and the new `CapabilityChainShapeTests` class gives solid targeted coverage. The multi-level `ValidateChainShapeAndResolveParentAsync` validator catches all four documented non-spec shapes. Good work.

Three things worth addressing before merge:

---

### 1. Inconsistent null-handling in `examples/ZcapLd.Examples/InMemoryDidProvider.cs` `RegisterRoot`

```csharp
// examples version — silently no-ops
public Capability RegisterRoot(Capability root)
{
    if (root is not null && !string.IsNullOrEmpty(root.Id))
        _rootStore[root.Id] = root;
    return root!;
}
```

The test-helper version (`tests/.../Helpers/InMemoryDidProvider.cs`) correctly throws `ArgumentNullException.ThrowIfNull(root)`. The examples version silently swallows a null or id-less root, returning `null!`. A caller following the examples would only see a confusing "root unavailable, verify fails closed" error at verification time, with no indication that registration was silently skipped. The examples code should throw (or at minimum match the guard in the test helper), since examples are the primary learning surface.

---

### 2. Exception-type inconsistency: `ExtractEntryId` throws `InvalidOperationException`

`CapabilityService.BuildCapabilityChain` throws `InvalidOperationException` when a parent's chain entry has no extractable id. The rest of the library raises `CapabilityValidationException` for validation failures. While this only surfaces at delegation time (which is correct — fail fast), the mixed exception types complicate caller error-handling. Consider throwing `CapabilityValidationException` (or a subclass) here for uniformity.

---

### 3. No negative test for `urn:zcap:root:` target-binding substitution hardening

`ResolveRootCapabilityAsync` has a security guard that decodes the `urn:zcap:root:<encodedTarget>` suffix and validates it against `root.InvocationTarget` to prevent root substitution attacks. This is the right thing to do, but it has no dedicated negative-path test — an attacker trying to pass a root with a matching id but a different `invocationTarget` would be rejected, but there's no test asserting that rejection. Given this is the primary substitution-prevention mechanism, it deserves explicit coverage, e.g.:

```csharp
[Fact]
public async Task Verify_Rejects_SubstitutedRoot_WithMatchingIdButDifferentTarget()
{
    var root = await RootAsync("https://example.org/real-target");
    // Craft a root with the same urn:zcap:root: id but a different invocation target
    var substituted = root with { InvocationTarget = "https://attacker.example/other-target" };
    var d1 = await DelegateAsync(root, ...);
    var result = await _verifier.VerifyCapabilityChainAsync(d1, substituted);
    result.Should().BeFalse();
}
```

---

### Minor note: merge conflict

The PR currently has `mergeable_state: dirty` — there is a conflict with `main` that needs to be resolved before merging.

---

Overall the design and implementation quality here are high. Items 1 and 2 are low-risk cleanups; item 3 is security-adjacent and worth adding before this ships as 3.0.0.

---

PR #98, opened 2026-06-05
feat(#70): structured VerificationResult failure-reason channel for verify methods

### LINKED ISSUE #70: [L5] Add a structured VerificationResult failure-reason channel to verify methods

moisesja:
## Summary

Closes #70 ([L5]). Every `IVerificationService` verify method returned a bare `Task<bool>`, so the distinct failure modes — revoked, expired, bad signature, broken chain, unauthorized controller, caveat failure, replay, chain-too-long — all collapsed to `false` with no programmatic way to tell them apart. Downstream consumers had to re-derive the reason (e.g. `net-wallet-sdk` reconstructed chain ids to recover `Revoked` vs `ChainInvalid`).

This adds a structured result channel **additively**, keeping the `bool` API fully backward-compatible.

## What changed

- **New types** (`Models/`): `VerificationOutcome` enum + `VerificationResult` record (`Outcome`, `Message?`, `IsValid`, `Valid`, `Fail(...)`).
- **`IVerificationService`**: 7 new `...DetailedAsync` siblings returning `Task<VerificationResult>` — proof (×2), invocation (×3), chain (×2). The existing `Task<bool>` methods are now thin wrappers: `(await ...DetailedAsync(...)).IsValid`.
- **`VerificationService`**: threaded `VerificationResult` through the internal verify helpers so each early `return false` carries a specific outcome at the orchestration layer (unambiguous — each check is already a separate statement). Revocation stays `bool` (control-plane, out of #70 scope; the only interface implementer is `VerificationService`, so nothing downstream breaks).
- **`CouldNotVerify`**: explicitly distinguishes a config/transient/infrastructure fault ("couldn't check") from a provably-invalid capability — surfacing the **M7 / #64** distinction at the API.
- **Logging gap (part B) closed**: a single `LogDenial` choke point on the public boundary logs **every** denial once at Debug (attacker-drivable → can't flood the Warning channel), not just the exception path. Previously the structurally-valid denials (revoked/expired/replay/…) returned `false` silently; only `#64`'s exception path was logged.

## Usage

```csharp
var result = await verificationService.VerifyInvocationDetailedAsync(invocation, capability);
if (!result.IsValid)
{
    // result.Outcome: Revoked / Expired / InvalidSignature / UnauthorizedController / Replayed /
    // CaveatFailed / ActionNotAllowed / InvalidTarget / ChainTooLong / MalformedCapability / CouldNotVerify
    log.Warn("denied: {Outcome} — {Message}", result.Outcome, result.Message);
}
// the bool method is unchanged and == result.IsValid
```

## Tests

- **16 new tests** (`VerificationResultTests.cs`) asserting each distinct outcome + bool parity + the null-arg throw contract.
- Full suite green: **Core 458 + AspNetCore 6, 0 failures**; build **0 errors / 0 warnings**.
- All 227 pre-existing Services/Verification tests pass unchanged (bool behavior preserved, fail-closed intact).

## Docs

ARCHITECTURE.md, README.md, AGENTS.md updated; plan + review in `tasks/`.

## Notes for reviewers

- `VerifyDelegationProofAsync` aggregates a multi-proof set by returning the last specific failure (exact in the common single-proof case); its per-proof catch classifies expected validation exceptions → `InvalidDelegation`, unexpected → `CouldNotVerify`, mirroring `LogFailedClosed`'s severity split.
- The `#64` logging work (closed separately) delivered only the *minimum* (exception-path logging); this PR delivers the *primary* structured-result ask and completes part B across all denial paths.


moisesja (commented):
## Review: `feat(#70)` — Structured `VerificationResult` failure-reason channel

CI green (both jobs). 458 Core + 6 AspNetCore tests pass. Full code review below.

---

### What's here

| New | Changed |
|---|---|
| `VerificationOutcome` enum (13 values) | `IVerificationService` +7 `…DetailedAsync` overloads |
| `VerificationResult` sealed record | `VerificationService` internal helpers return `VerificationResult` |
| `VerificationResultTests.cs` (16 tests) | `Task<bool>` methods are thin `(await …Detailed).IsValid` wrappers |
| `LogDenial` choke point | ARCHITECTURE / README / AGENTS docs updated |

---

### Design — looks good

**Additive, backward-compatible.** The existing `Task<bool>` surface is unchanged in semantics; callers who don't need the reason see zero diff.

**`VerificationResult.Valid` static singleton** avoids an allocation on every success. The `sealed record` + `required init` shape means the shared instance is immutable — can't be accidentally mutated by a caller. ✓

**`VerificationOutcome.Valid = 0`** is the right default: the zero-value is the safe/success shape. Because `VerificationResult` uses `required`, you still can't default-construct it without explicitly setting `Outcome`, so there's no accidental "default = valid" footgun. ✓

**`LogDenial` choke point** — all structurally-valid denials now log once at Debug regardless of whether the caller used the `bool` or `Detailed` path. Debug severity for attacker-drivable branches is the right trade-off (hostile client can't flood the Warning channel). `CouldNotVerify` is correctly excluded here because `LogFailedClosed` already handled it with type-aware severity at the fault site — no double-logging. ✓

**Multi-proof `lastFailure` aggregation** in `VerifyDelegationProofAsync` returns the last specific failure reason when every proof in the set fails. Exact for the common single-proof case, and the behaviour is documented in the PR notes and in the code comment. Acceptable. ✓

---

### One minor note (no action required)

`VerifyCapabilityProofAsync(Capability, Capability)` changed from returning a task directly to `async`/`await`. The added state-machine overhead is negligible here (this sits on top of async crypto I/O), but worth knowing if this method shows up in an allocation trace later.

---

### Summary

Clean, well-scoped implementation of #70. The structured result channel is additive, fail-closed semantics are preserved end-to-end, the `CouldNotVerify` vs "provably-invalid" distinction is surfaced correctly at the API boundary, and the 16 new tests exercise every named outcome. Nothing blocks merge.

---

Issue #99, opened 2026-06-05
[L7] VerifySingleDelegationProofAsync has no proof.created freshness/consistency check


moisesja:
## Summary
Delegation-proof verification (`VerifySingleDelegationProofAsync`) checks `capability.expires` but never inspects the delegation `proof.created` timestamp. A delegation proof can therefore claim any `created` value — including one **in the future**, or one **after the capability's own `expires`**, or after the parent's validity window — and still verify. Follow-up to #71, which added `created` freshness to the *invocation* and *revocation* paths.

## Why this is NOT the same check as #71
Invocations are ephemeral and replay-sensitive, so #71 rejects `now − created > nonceWindow` (a staleness lower-bound). Delegation proofs are **durable** — a capability delegated months ago is legitimately still valid until it `expires`. Reusing #71's staleness bound here would wrongly reject every capability older than the nonce window. The correct delegation check is narrower:

- `created` MUST be present and parseable (fail closed on null/unparseable);
- `created` MUST NOT be in the future beyond a small clock-skew tolerance;
- `created` MUST be `<= expires` when `expires` is present (temporal consistency);
- SHOULD be consistent with the parent's window (parent `created <=` child `created`, child window within parent window) — design to confirm.

**No staleness lower-bound.**

## Location
`src/ZcapLd.Core/Services/VerificationService.cs` — `VerifySingleDelegationProofAsync` (expiry check ~lines 415–418; chain-path expiry ~lines 719–720; parent/child attenuation ~lines 946–952). None read `proof.CreatedAt`.

## Recommended fix
Add a delegation-specific freshness/consistency check (distinct from #71's `GetFreshProofCreatedUtc`) in the single-proof and chain delegation paths; fail closed on a missing, future-beyond-skew, or post-`expires` `created`. Add compliance + unit tests (future-dated delegation rejected; `created > expires` rejected; old-but-unexpired delegation still verifies).

## Impact
**Low.** Not a spec MUST; `expires` already bounds the outer validity. Tightens temporal soundness of the delegation chain. Found alongside #71 (adversarial audit follow-up).

PR #100, opened 2026-06-05
fix(#71): enforce proof.created freshness in invocation & revocation verification

### LINKED ISSUE #71: [L6] VerifyInvocationAsync has no Proof.Created freshness check

moisesja:
## Summary

Closes #71. Enforces `proof.created` freshness on the invocation and signed-revocation verify paths, bounding replay independently of nonce-store eviction.

Previously `VerifyInvocationAsync` / `RevokeCapabilityAsync` verified the signature and tracked the request id in the nonce store but **never inspected the signed `created` timestamp**, so replay was bounded only by the nonce store's eviction TTL (`_nonceWindow`, default 5 min). Once the entry evicted, an identical captured invocation re-verified; with the opt-out `NullNonceStore` replay was unbounded.

This re-applies the abandoned PR #90 fix **by hand** against current `main` — PR #90 was the tip of a 7-commit stack whose other six fixes already merged independently, so merging that branch double-applied them. Only the #71 delta was genuinely missing.

## Changes

**Production (`src/ZcapLd.Core/`):**
- `VerificationService` — new `FreshnessClockSkew` constant + shared `GetFreshProofCreatedUtc` helper. Rejects `created` that is missing/unparseable, future-dated beyond a 1-min clock skew, or older than `_nonceWindow`. Reusing the nonce window as the staleness bound gives the invariant: *anything still evictable from the store is already too stale to pass freshness*.
- **Invocation path** (`VerifyInvocationCoreAsync`) — freshness gate as step 2c, before the expensive signature/authorization work; `InvocationTime` now sourced from the validated signed `created` so time-based caveats evaluate against the invoker's signed time.
- **Revocation path** (`RevokeCapabilityInternalAsync`) — same gate (stays `bool` — control-plane).
- New `VerificationOutcome.StaleProof` detailed-API reason.
- `SigningService` — `createdOverride` determinism **overloads** (kept off `ISigningService` — no interface break).

**Tests:** new `VerificationServiceFreshnessTests.cs` — 8 facts asserting the exact `StaleProof` outcome: fresh / stale / future-beyond-skew / future-within-skew / null-created / stale-under-`NullNonceStore` (invocation), plus fresh / stale (revocation). **All 472 tests pass** (466 Core + 6 AspNetCore), no regressions.

**Docs:** CHANGELOG (Fixed + Added) and AGENTS.md updated.

## Scope decisions

- **Revocation path covered** in addition to invocation (identical gap, line 1372).
- **`InvocationContext.InvocationTime` sourced from signed `created`** (bounded by the freshness window) — the issue's optional suggestion.
- **Delegation-proof freshness is out of scope** — it has *distinct* semantics (delegations are durable; reusing this staleness bound would wrongly reject every capability older than the nonce window). Tracked separately in #99.

## Verification

```
dotnet build ZcapLd.sln   # clean
dotnet test  ZcapLd.sln   # 472 passed, 0 failed
```


moisesja (commented):
## Review — fix(#71): proof.created freshness

Solid, well-scoped security fix. The core invariant ("anything still evictable from the nonce store is already too stale to pass freshness") is sound and elegantly avoids a second tuneable constant. All three rejection cases (missing, future-beyond-skew, older-than-window) are covered, tests are well-isolated via the `createdOverride` determinism overload, and the revocation path correctly stays `bool` to match its existing convention.

Would approve — blocked only by self-review. Two minor non-blocking observations:

---

**1. Comment step-numbering is inverted in `VerifyInvocationCoreAsync`**

The freshness gate is labelled `// 2c.` but appears _before_ the block labelled `// 2a.`. A reader following the numbers will be confused. Renumber the gate to fit sequentially where it sits (e.g. `// 2b.`), or swap the comment ordering.

---

**2. `DefaultNonceWindow` doc doesn't surface its dual role**

`_nonceWindow` now governs both nonce eviction *and* the proof-created staleness bound. A consumer who widens it to 1 hour to accommodate slow agents will implicitly accept invocations signed up to 1 hour ago — an interplay invisible from the current XML doc on `DefaultNonceWindow`. A one-liner note would make this explicit:

> _"Also serves as the upper staleness bound for `proof.created` freshness checking (Issue #71); widening this window extends both the replay window and the acceptable signing age."_

---

Neither item affects correctness or security. 472 tests passing, delegation-proof freshness correctly scoped out to #99. LGTM.

---

moisesja:
Addressed in `974350f`:

- **Review nit 1 (step numbering):** renumbered the invocation-verify comments so they read monotonically — `2` (purpose) → `2a` (freshness) → `2b` (capability reference) → `2c` (proof-field consistency). The freshness gate was labelled `2c` but sat before `2a`.
- **Review nit 2 (`DefaultNonceWindow` dual role):** added the suggested note to its XML doc — it now states the window also serves as the `proof.created` staleness bound, so widening it extends both the replay window and the acceptable signing age.
- **Configurable clock skew:** the future-dated tolerance is now per-instance — new public `DefaultFreshnessClockSkew` (1 min) + optional `freshnessClockSkew` constructor parameter on the `INonceStore`-bearing constructors, mirroring `nonceWindow`. New test `VerifyInvocation_FutureProof_HonorsConfiguredClockSkew` proves a +90s proof is rejected (`StaleProof`) under the default 1-min skew and accepted (`Valid`) under a configured 2-min skew.

473 tests passing (467 Core + 6 AspNetCore).

PR #101, opened 2026-06-05
docs+test: document root drops expires/allowedActions/caveats (#72)

### LINKED ISSUE #72: [L7] CreateRootCapabilityAsync silently drops expires/allowedActions/caveats

moisesja:
Closes #72.

`CreateRootCapabilityAsync` accepts `expires` / `allowedActions` / `caveats` but silently drops all three on root capabilities — which is **correct** per W3C ZCAP-LD (a root carries none; a root that carries `expires` is actively rejected by `ValidateCapabilityAsync`). The wart the issue describes is the misleading signature, not the drop.

This lands the non-breaking fix the issue recommends — **document the parameters as ignored** (removing them is source-breaking; throwing would break every root-creation call site). It pairs with #66, which already aligned the `allowedActions` signature and its doc.

## Changes
- **`ICapabilityService.cs`** — rewrite the `expires` and `caveats` XML param docs to state they are ignored / not honored for roots (the `allowedActions` doc was already corrected in #66).
- **`CapabilityServiceTests.cs`** — add `CreateRootCapability_DropsExpiresAllowedActionsAndCaveats`, a regression test that locks the drop so a future change does not start honoring these on roots (the issue explicitly warns: do **not** honor `expires` on roots).
- **`CHANGELOG.md`** — entry under Fixed.

Behavior is unchanged — this is docs + test only.

## Provenance
Cherry-picked (`git cherry-pick -x`) from the previously-unmerged branch `72-root-dropped-params-doc` (commit `c8b8ced`), rebased onto current `main`. The only conflict was in `CHANGELOG.md`; resolved by keeping `main`'s richer #71/#51/#69/#68/#67 entries and prepending the new #72 line.

## Verification
`dotnet test --filter FullyQualifiedName~CapabilityServiceTests` → **34 passed, 0 failed**, including the new regression test.


moisesja (commented):
Clean, focused PR — this is exactly what a "document + lock" change should look like.

**What's here:**
- `ICapabilityService.cs` — `expires` and `caveats` XML doc params now clearly state they are **ignored** on root capabilities, with the spec rationale (`ValidateCapabilityAsync` actively rejects a root that carries `expires`) and issue reference. Much better than the old one-liners.
- `CapabilityServiceTests.cs` — new regression test locks the intentional drop: all three ignored params are supplied, all three come back `null` on the root, and the clean root validates. The inline comment explaining *why* the drop is correct (not a bug) is a good touch that will prevent future "let me fix this" PRs.
- `CHANGELOG.md` — entry is appropriately detailed.

**Spot-checked:** `ExpirationCaveat` exists in `src/ZcapLd.Core/Models/Caveat.cs`, so the test compiles cleanly.

No concerns. LGTM.

---

PR #102, opened 2026-06-05
feat(#73): add opt-in verifier-side 3-month expiration ceiling

### LINKED ISSUE #73: [L8] Verifier-side 3-month expiration SHOULD is absent

moisesja:
Closes #73.

## What #73 was

The W3C ZCAP-LD verifier-side **SHOULD** — *"a verifier SHOULD ensure that an invoked delegated zcap does not have an expiration date-time more than three months in the future"* — was absent on the verify path. The MUST "not expired" check was present, but the 3-month ceiling lived (wrongly) as a create-time hard `throw`, which **#61** removed. This PR adds the check where the spec places it: on the **verifier**, at **verification time**, behind a policy flag (it is a SHOULD). It is the companion work item to #61.

> Note: a prior cumulative PR (#92) implemented this against an older `main`, but was closed without merging. This re-applies the reviewer-approved design onto current `main` (post-#70 `VerificationResult`, post-#50/#51/#65/#71 constructor), with two hardening fixes from adversarial review.

## Implementation

- **`VerificationPolicy`** (new, `ZcapLd.Core.Services`) — opt-in `EnforceMaxDelegationExpiration` (default `false`) + configurable `MaxDelegationExpirationMonths` (default `3`).
- **`VerificationService`** — `policy:` parameter on the full constructor; the ceiling is enforced per delegated chain link in `VerifyBuiltChainAsync` (covers the invocation **and** chain-verification paths), measured at `DateTime.UtcNow`.
- **`VerificationOutcome.ExpirationTooFarInFuture`** — new structured reason (distinct from `Expired`, the past-expiry MUST), integrating with the #70 detailed-result channel.
- **DI** — `AddZcapServices()` picks up a registered `VerificationPolicy` (`services.AddSingleton(new VerificationPolicy { EnforceMaxDelegationExpiration = true })` before `AddZcapServices`).
- Docs updated: CHANGELOG, ARCHITECTURE, spec-requirements checklist, security summary, compliance matrix. The misnamed SHOULD-04 compliance test now genuinely exercises the verifier.

## Hardening from adversarial review

Two skeptic agents probed the change; both findings were real and are fixed:

1. **Revocation must stay possible.** `VerifyBuiltChainAsync` is also reached by the revocation-authorization path. An `applyExpirationCeiling` flag lets `IsRevokerAuthorizedAsync` pass `false`, so a long-lived delegation remains **revocable** — refusing to authorize its removal would be exactly backwards (revocation is the storage-burden mitigation the ceiling exists for).
2. **Omit-`expires` bypass closed.** A delegated zcap with **no** `expires` is unbounded (strictly worse for storage burden) yet slipped through. Under the policy it is now rejected too. The bound also applies to **every** delegated link (not just the invoked leaf), since the verifier retains each link's revocation until it expires.

Default-off means **zero behavior change** for existing callers.

## Verification

- Clean build (0 warnings).
- **480 tests pass** (474 Core + 6 AspNetCore), incl. 8 new/updated: policy off verifies / on rejects (invocation + chain), within-ceiling passes, configurable months, long-lived intermediate rejected, missing-`expires` rejected, and revocation still succeeds under the strict policy.


moisesja (commented):
Solid implementation of a subtle spec requirement. The design decisions are all sound and the adversarial-review findings were well-addressed. A few minor observations — none are blockers.

---

**What's good**

- **Opt-in by default is exactly right.** A SHOULD that can reject legitimately long-lived delegations must not change behavior for existing callers. `VerificationPolicy.Default` / `EnforceMaxDelegationExpiration = false` handles this cleanly with zero behavior change for existing deployments.
- **Revocation bypass (`applyExpirationCeiling: false`)** is the critical correctness point — refusing to authorize removal of a long-lived delegation would be exactly backwards. It's named clearly, documented in both the method XML-doc and inline comment, and regression-guarded by `RevokeCapability_ExpirationBeyondCeiling_PolicyOn_StillSucceeds`.
- **Missing `expires` → rejected under policy.** The omit-bypass was caught in adversarial review and correctly closed: an unbounded delegation is strictly worse for storage burden than any fixed expiration, so treating `null` as "infinitely far out" is the right call.
- **Per-link enforcement (not leaf-only)** is well-reasoned: the verifier stores each link's revocation until it expires, and under attenuation the ancestors are the longest-lived. The intermediate-link test (`VerifyChain_LongLivedIntermediate_PolicyOn_Rejected`) pins this behavior.
- **`sealed record VerificationPolicy`** — value semantics, immutable, correct type choice.
- **Test coverage** is comprehensive: policy off/on for both the invocation and chain-verify paths, configurable ceiling months, intermediate link, missing expires, and the revocation bypass. The compliance test now genuinely exercises the verifier path rather than just create-time behaviour.

---

**Minor nits (non-blocking)**

1. **`DateTime.UtcNow` read inside the per-link loop** — the comparison `childExpiresAt.Value > DateTime.UtcNow.AddMonths(...)` is called once per delegated link. For a max-10-link chain this is harmless (sub-millisecond total), but capturing `var ceilingCutoff = DateTime.UtcNow.AddMonths(_policy.MaxDelegationExpirationMonths)` before the loop would be slightly cleaner and consistent with how the freshness helpers snapshot time once.

2. **`MaxDelegationExpirationMonths ≤ 0` is documented but untested.** The XML-doc says "a value ≤ 0 rejects every future-dated delegation" — worth a single guard or a brief test to pin that edge, since a misconfigured value of 0 would silently block all delegated invocations with no obvious error message.

Neither warrants holding the PR. The implementation is correct, the documentation is thorough, and the test suite gives strong confidence in all the behavioral boundaries.

---

moisesja:
Thanks for the review. Both nits addressed in 3f64be5:

**Nit 1 — `DateTime.UtcNow` in the per-link loop.** Hoisted: the cutoff is now snapshotted once before the loop, and *only* when the policy is active (so `AddMonths` is never evaluated when the ceiling is off or on the revocation path), consistent with how the `proof.created` freshness helpers snapshot time once:

```csharp
DateTime? expirationCeilingCutoff = (applyExpirationCeiling && _policy.EnforceMaxDelegationExpiration)
    ? DateTime.UtcNow.AddMonths(_policy.MaxDelegationExpirationMonths)
    : null;
// ... in the loop:
if (expirationCeilingCutoff is { } ceiling) { /* missing-expires + > ceiling checks */ }
```
One cutoff for the whole chain is also marginally more correct (every link compared against the same instant) and can never loosen the bound.

**Nit 2 — `MaxDelegationExpirationMonths <= 0` foot-gun.** Now fails fast: the `init` setter throws `ArgumentOutOfRangeException` for any value `< 1`, so a misconfigured `0` surfaces at construction instead of silently rejecting every delegated invocation. `Default` (3) is unaffected. Pinned by `Policy_MaxDelegationExpirationMonthsBelowOne_Throws` (0 and -1) and `Policy_MaxDelegationExpirationMonthsOfOne_IsAccepted`.

I re-ran the adversarial pass on both changes: it confirmed the hoist is a strict, equivalence-preserving improvement (never loosens the ceiling; missing-`expires` rejection unchanged) and the guard is unbypassable (`Default` doesn't throw at type-init; `with`-expressions/object-initializers all throw; the record copy-ctor can't produce `<1`). 483 tests pass.

PR #103, opened 2026-06-05
fix(#74): use ordinal StartsWith in invocation-target prefix check

### LINKED ISSUE #74: [L9] Culture-sensitive StartsWith in invocation-target prefix check

moisesja:
Closes #74.

## Summary
`VerificationService.IsValidInvocationTarget` tested the prefix with the **culture-sensitive** `string.StartsWith(string)` overload, while the surrounding logic is all **ordinal** — the `==` exact-match, the following `Substring(capabilityTarget.Length)` index math, and the `StartsWith(char)` suffix checks. With an ignorable Unicode code point (e.g. U+00AD soft hyphen) the culture-sensitive comparison disagrees on length; when the capability target is ordinally *longer* than the invocation target it returns `true` and the subsequent `Substring` indexes past the end → `ArgumentOutOfRangeException`.

Both call sites (invocation verify, delegation attenuation) are fail-closed, and real targets are ASCII URIs, so impact is **low** — a latent correctness/consistency bug.

## Fix
```diff
-        if (invocationTarget.StartsWith(capabilityTarget))
+        if (invocationTarget.StartsWith(capabilityTarget, StringComparison.Ordinal))
```
For all real ASCII URIs ordinal and culture-sensitive agree, so there is **no behavior change** for legitimate input; only the pathological Unicode path is corrected and the `ArgumentOutOfRangeException` corner case removed.

`IsValidInvocationTarget` is promoted `private` → `internal` (the repo already has `InternalsVisibleTo ZcapLd.Core.Tests`) so the prefix logic can be unit-tested directly — the fail-closed callers swallow the bug into a bare `false`, making a black-box test unable to distinguish the fix from the bug.

## Tests (8 new)
- 4 valid-prefix cases (exact, path suffix, query suffix, extra `&` param) → `true`
- 3 invalid-prefix cases (different host, suffix without a delimiter, ignorable char breaking the ordinal prefix) → `false`
- 1 guard `Fact`: capability target ordinally longer via a trailing U+00AD must return `false` **without throwing**

**Proven by ablation:** with the fix reverted to the culture-sensitive overload (and the test project rebuilt so it links the reverted `ZcapLd.Core` — not run with `--no-build`, which would link the stale fixed copy), the guard `Fact` fails with the exact `ArgumentOutOfRangeException: startIndex cannot be larger than length of string` from the issue. With the fix, all 8 pass; the 7 non-throwing controls pass on both versions.

## Verification
- Full suite green: **486 Core + 6 AspNetCore**, 0 failures.
- An out-of-band probe (.NET 10 / macOS / en-US / ICU) confirmed the divergence is real on this platform for U+00AD and other ignorable code points.

No public API, wire-format, or documentation-surface change (internal behavior only).

### DIFF
--- a/src/ZcapLd.Core/Services/VerificationService.cs
+++ b/src/ZcapLd.Core/Services/VerificationService.cs
@@ -1198,7 +1198,9 @@ public class VerificationService : IVerificationService
     /// Per spec: suffix MUST start with slash or question mark if no query exists,
     /// or ampersand when the capability target already contains a query.
     /// </summary>
-    private bool IsValidInvocationTarget(string invocationTarget, string capabilityTarget)
+    /// <remarks>Internal (not private) so the ordinal prefix matching can be unit-tested
+    /// directly; the fail-closed call sites make a black-box test indistinguishable (Issue #74).</remarks>
+    internal bool IsValidInvocationTarget(string invocationTarget, string capabilityTarget)
     {
         if (string.IsNullOrEmpty(invocationTarget) || string.IsNullOrEmpty(capabilityTarget))
             return false;
@@ -1207,8 +1209,12 @@ public class VerificationService : IVerificationService
         if (invocationTarget == capabilityTarget)
             return true;
 
-        // Check if invocationTarget is a valid prefix extension
-        if (invocationTarget.StartsWith(capabilityTarget))
+        // Check if invocationTarget is a valid prefix extension.
+        // Ordinal comparison keeps this consistent with the ordinal Substring index math
+        // below (and the ordinal `==`/char checks around it); a culture-sensitive overload
+        // can treat ignorable Unicode code points as zero-width, disagreeing on length and
+        // throwing ArgumentOutOfRangeException in the subsequent Substring (Issue #74).
+        if (invocationTarget.StartsWith(capabilityTarget, StringComparison.Ordinal))
         {
             var suffix = invocationTarget.Substring(capabilityTarget.Length);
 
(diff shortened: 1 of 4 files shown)


moisesja (commented):
## Code Review: fix(#74) — ordinal `StartsWith` in invocation-target prefix check

The fix is correct and well-scoped. The root cause is real: the culture-sensitive `string.StartsWith(string)` overload treats ignorable Unicode code points (U+00AD, U+200B, etc.) as zero-width, so `capabilityTarget`'s ordinal length can exceed `invocationTarget`'s, causing the subsequent `Substring(capabilityTarget.Length)` to throw `ArgumentOutOfRangeException`. Switching to `StringComparison.Ordinal` is exactly right — no behavior change for legitimate ASCII URIs, and the pathological path is closed cleanly.

The `private` → `internal` promotion is a pragmatic necessity: the fail-closed callers catch the exception and return `false`, making a black-box test unable to distinguish the bug from the fix.

**Tests** are thorough:
- Valid-prefix theory covers exact match, path extension, query extension, and multi-param query
- Invalid-prefix theory covers wrong host, delimiter-less suffix, and the specific ignorable-char case
- The guard `Fact` pins the no-throw contract directly against the `ArgumentOutOfRangeException` failure mode
- Ablation was verified by rebuilding the **test project** (not just the library), correctly avoiding the stale-DLL false-green trap documented in `tasks/lessons.md`

**Two minor style points** (non-blocking):

1. **Multi-line in-code comment block.** The 4-line `// Ordinal comparison keeps this consistent…` block is longer than the project's "one short line max" guideline. The WHY is genuinely non-obvious, but it could be trimmed to one or two lines without losing the invariant.

2. **Issue reference in `<remarks>`.** `(Issue #74)` in the XML remarks on `IsValidInvocationTarget` will rot once the issue is closed. The functional explanation is what belongs there; the tracker citation belongs in the PR description (where it already lives).

Neither point affects correctness or security. The logic, test design, and captured lessons are all solid. ✅

---

PR #104, opened 2026-06-06
feat(#99): delegation-proof created soundness check

### LINKED ISSUE #99: [L7] VerifySingleDelegationProofAsync has no proof.created freshness/consistency check

moisesja:
Closes #99.

## Problem
Delegation-proof verification (`VerifySingleDelegationProofAsync`) checked `capability.expires` but **never inspected the delegation `proof.created`**, so a delegation proof could claim any `created` — future-dated, etc. — and still verify. Follow-up to #71, which added `created` freshness to the *invocation/revocation* paths. Severity: low (hardening; not a spec MUST; `expires` already bounds outer validity).

## Approach
One helper, `CheckDelegationProofCreated`, called at the **top of `VerifySingleDelegationProofAsync`** — the single funnel every path shares (standalone, chain, invocation, revocation-auth) — ordered **before** the signature gate (mirrors #71, keeps the negative paths observable without a valid signature).

| Case | Behavior |
|---|---|
| `created` future-dated beyond `_freshnessClockSkew` (default 1 min) | **always rejected** |
| `created` present-but-unparseable | **always rejected** — reads raw `Proof.Created` and parses defensively (catch `FormatException` only), so the reason is the precise `InvalidProofTime`, not the per-proof catch's `InvalidDelegation` |
| `created` missing/empty | rejected **only** under opt-in `VerificationPolicy.RequireDelegationProofCreated` (default `false`) — pre-existing/cross-stack delegations may legitimately omit it |

**No staleness lower-bound** — a durable delegation signed months ago must verify until it `expires` (the #71↔#99 distinction); does **not** reuse `GetFreshProofCreatedUtc`.

Threaded via `applyCreatedCheck` on `VerifyDelegationProofAsync` / `VerifySingleDelegationProofAsync` / `VerifyBuiltChainAsync`; the **revocation-authorization path passes `false`** so a malformed-`created` delegation stays revocable (mirrors the #73 ceiling carve-out).

### Scope decisions
- **Dropped** the issue's `created <= expires` rule: it is redundant (every `created > expires` case is already caught by the existing `Expired` check or the future-dated check) and false-reject-prone at the clock-skew boundary.
- **Deferred** the parent/child window-consistency SHOULD.

## API additions
- `VerificationOutcome.InvalidProofTime` (distinct from `StaleProof`/`Expired`)
- `VerificationPolicy.RequireDelegationProofCreated` (default `false`)
- `SignCapabilityAsync(.., createdOverride)` determinism overload on the concrete `SigningService` (test seam; not on `ISigningService`)

## Tests
- `VerificationServiceDelegationFreshnessTests` (8): future-beyond-skew rejected (chain + standalone), within-skew verifies, **90-day-old-but-unexpired still verifies** (no-lower-bound regression guard), unparseable always rejected, missing rejected under policy / bypassed by default, future-dated delegation still revocable
- `VerificationResultTests`: `InvalidProofTime` case
- `NormativeIntegrationComplianceTests`: SHOULD-08

**502/502 tests pass** (496 Core + 6 AspNetCore). Adversarial review of the diff found the fix sound with no must-fix issues (funnel completeness, reject-only gate, `FormatException`-only catch empirically justified, revocation carve-out uniquely plumbed, no false-reject regression).

Docs updated: `AGENTS.md`, `ARCHITECTURE.md`.


moisesja (commented):
Clean, well-scoped security hardening. All the key design decisions check out:

**Design correctness**
- `CheckDelegationProofCreated` is correctly ordered *before* the signature gate — cheap local check, negative paths observable without resolver I/O, mirrors the #71 pattern exactly.
- Reading raw `Proof.Created` (not `Proof.CreatedAt`, which throws) is the right call — ensures the unparseable case surfaces as `InvalidProofTime` rather than being reclassified as `InvalidDelegation` by the per-proof catch.
- `FormatException`-only catch is precise and justified — no swallowing of genuine faults.
- **No staleness lower-bound** is exactly right for durable delegation proofs. The regression guard (`VerifyChain_OldButUnexpiredDelegation_StillVerifies`) locks this in.
- Revocation carve-out (`applyCreatedCheck: false`) mirrors the #73 expiration-ceiling pattern consistently.
- Dropping the `created <= expires` rule is correct — it is redundant and false-reject-prone at the clock-skew boundary.
- `RequireDelegationProofCreated` default `false` preserves backwards compat with pre-existing/cross-stack delegations.

**Threading**
- `applyCreatedCheck` is plumbed correctly through `VerifyDelegationProofAsync` → `VerifySingleDelegationProofAsync` → `VerifyBuiltChainAsync`, and `IsRevokerAuthorizedAsync` passes `false` as expected.

**Tests**
- 10 new tests cover all meaningful cases: future-dated (chain + standalone), within-skew passes, 90-day-old durable delegation still verifies (the critical regression guard), unparseable always-rejected, missing rejected under policy / passes by default (with correct `InvalidSignature` outcome proving the gate was bypassed, not just silent), and the revocation carve-out.
- SHOULD-08 compliance test is a solid addition.

**One acknowledged nit** (pre-existing scope, not a regression): the standalone `VerifyCapabilityProofAsync` single-link path checks only the leaf proof's `created`; deep chains on that path skip intermediate links. Chain/invocation paths are unaffected. Fine to defer.

502/502 tests passing. LGTM — ready to merge once CI is green.

---

moisesja:
Thanks for the review. Addressed the acknowledged nit in 2fcbede.

**Nit**: the standalone `VerifyCapabilityProofAsync` single-link path checks only the leaf proof's `created`; on a deep delegation an intermediate ancestor's `created` is not gated there. (Pre-existing single-link scope, not a regression.)

Rather than expand the single-link method into full-chain verification — that would change its documented contract, duplicate `VerifyCapabilityChainAsync`, and be internally inconsistent since intermediate *signatures* aren't verified on that path either — I made the boundary **explicit and intentional**:

- **Docs** (`IVerificationService`): the single-link summary now lists the leaf's `created` (#99), and the "not a full authorization gate" caveat names `created` alongside signature/attenuation for ancestors, spelling out that on a deep delegation only the leaf link's `created` is checked here — an intermediate's `created`, like its signature, is gated only by the full-chain path.
- **Test** (`VerifyCapabilityProof_DeepChain_GatesLeafOnly_ChainPathGatesEveryLink`): a `root → mid → leaf` chain with a future-dated **intermediate** `created` verifies on the single-link path (`Valid`) but is rejected on the chain path (`InvalidProofTime`), pinning the boundary so it stays intentional.

503/503 tests pass.

Issue #108, opened 2026-06-14
Architecture: deduplicate crypto/proof onto NetCrypto + NetDid 2.0.0 + DataProofs (and the Ed25519Signature2020 delegation path)


moisesja:
## Summary

Evaluate and execute deduplication of `ZcapLd.Core`'s home-grown cryptography and linked-data-proof code in favor of the now-published composable stack — **NetCrypto 1.0.0**, **NetDid 2.0.0** (now built on NetCrypto), **NetCid 1.6.0**, and the **DataProofsDotnet.\*** family — so zcap delegates as much as possible to upstream libraries it already owns, while preserving wire compatibility and the ZCAP-LD security model.

This issue also answers a specific design question raised during evaluation: **why the DataProofs cryptosuites cannot verify zcap's current `Ed25519Signature2020` proofs out of the box, and exactly how to make DataProofs host that convention so zcap can still delegate the proof pipeline upstream without a wire break.**

**TL;DR recommendation:**
1. **Stage A (mandatory, stable, wire-compatible major):** upgrade to NetDid 2.0.0 and move zcap's raw crypto to NetCrypto in the same step (the crypto types left NetDid's namespace, so this is forced). One shared crypto stack. Existing capabilities still verify — no re-delegation.
2. **Stage B (preview-gated):** replace zcap's RDFC + JCS canonicalization with `DataProofsDotnet.Rdfc` / `NetCid` behind adapters.
3. **Maximal delegation (the answer to #2):** add a **legacy `Ed25519Signature2020` / `EcdsaSecp256r1Signature2019` `ICryptosuite`** upstream in DataProofs, then drive zcap sign/verify through `DataIntegrityProofPipeline` — deleting `ProofSigningPayloadBuilder` / `CryptoSuite` / `SignatureVerifier` and keeping only the ZCAP policy layer. No wire break.
4. **Always first:** land a **golden-vector regression harness** — the repo currently has zero hardcoded proof vectors, so any byte drift on a swap would ship undetected.

---

## 1. The new composable stack (verified against published packages)

| Package | Version | Role | Stability |
|---|---|---|---|
| `NetCrypto` | **1.0.0** | Consolidated crypto: `ISigner`, `IKeyStore`, `KeyPair`, `DefaultCryptoProvider`, `DefaultKeyGenerator`, `KeyType` (enum), `EcdsaSignatureFormat`, `JwkConverter`, bundled zkryptium BBS FFI | **stable** |
[...]

PR #109, opened 2026-06-15
Migrate crypto/proof to the composable stack (NetCrypto / DataProofs / NetCid) — 4.0.0 (#108)

moisesja:
## Summary

Migrates `zcap-dotnet` onto the composable foundation (issue #108): all cryptography, canonicalization, and proof machinery are delegated to the shared stack — **NetCrypto** (via **NetDid 2.0**), **DataProofs** (Core/Rdfc/**Legacy**), and **NetCid** — and zcap stops shipping its own. This is a **4.0.0** major release.

**Wire-compatible:** the `Ed25519Signature2020` / `EcdsaSecp256r1Signature2019` proof bytes are unchanged, so capabilities issued by earlier versions still verify — **no re-delegation**. It is source-breaking (consumers recompile; some types removed).

> Versioning note: `Directory.Build.props` is `4.0.0` (this branch's whole history is unreleased — latest published is `0.1.1`). The CHANGELOG `[4.0.0]` section subsumes the unreleased `[3.0.0]` section below it.

## Why

zcap is a ZCAP-LD domain library, not a crypto or proof-format library. The foundation now provides everything it needs, so zcap should consume it rather than maintain a parallel copy of RFC 8785, RDFC-1.0, the Data Integrity proof pipeline, and per-curve sign/verify. Net result: **−hand-rolled crypto, −dotNetRDF direct dependency, −the crypto-suite extension surface**; zcap keeps only ZCAP-LD semantics (capabilities, chains, caveats, invocation, revocation, authorization policy).

## How it was staged

Each stage is gated byte-for-byte by a **Phase 0 golden-vector harness** (deterministic `proofValue`, RDFC N-Quads + hash-concat payload, JCS null-skip) so wire compatibility is provable at every step:

1. **Phase 0** — golden-vector harness (`tests/.../Compliance/ProofGoldenVectorTests`, `CanonicalizationGoldenVectorTests`). The repo previously had **zero** hardcoded proof vectors — only in-process round-trips — so drift was invisible.
2. **Stage A** — raw crypto → **NetCrypto** via **NetDid 2.0** (which relocated its crypto to NetCrypto). One shared crypto stack.
3. **Stage B** — **RDFC-1.0** → `DataProofsDotnet.Rdfc` (drops the direct dotNetRDF reference); **JCS** → `NetCid.JcsCanonicalizer` (fixes a latent culture-sort/locale determinism bug).
4. **Stage C** — proof signing/verification → **DataProofs.Legacy** cryptosuites (`Ed25519Signature2020` / `EcdsaSecp256r1Signature2019`) via a `LegacyProofCrypto` engine + a `DidSignerAdapter` (`IDidSigner` → `NetCrypto.ISigner`). zcap keeps its proof model, chain walking, the #68 key-type binding, and all policy.
5. **Cleanup** — retired the now-dead crypto types and **removed the crypto-suite extension surface** entirely (zcap is not a crypto-extension point; see `docs/crypto-suite-extensibility-decision.md`).

## Dependency changes

**Before:** `dotNetRdf.Core` (direct) + `NetDid 1.3.1` (its own NSec/NBitcoin/Nethermind crypto).

**After:** `NetCrypto 1.1.0`, `NetDid.Core` / `NetDid.Method.Key 2.0.1`, `DataProofsDotnet.Core` / `.Rdfc` / `.Legacy 1.0.0`, `NetCid 1.6.0` — one shared crypto stack, dotNetRDF transitive via DataProofs.Rdfc.

## Breaking changes (all source-level; wire unchanged)

- `NetDid.Core.Crypto.*` types are now `NetCrypto.*` (NetDid 2.0).
- **Removed:** `ICryptoSuite`, `ICryptoSuiteProvider`, `CryptoSuite`, `CryptoSuiteProvider`, `IDocumentCanonicalizerProvider`, `DocumentCanonicalizerProvider`, `SignatureVerifier`, and `AddZcapCryptoSuite<T>()`.
- `SigningService` / `VerificationService` drop their suite-provider + canonicalizer-provider constructor parameters; canonicalization is now a `canonicalizationMethod` argument (`"JCS"` default / `"RDFC-1.0"`) or `AddZcapRdfcCanonicalization()`.
- `IDidSigner` is **unchanged** (consumer HSM/KMS integration unaffected).

Adding a new curve is now a foundation change (a NetCrypto primitive + a DataProofs cryptosuite, wired into the internal `ZcapSuiteCatalog` + `LegacyProofCrypto`) — never a zcap API.

## Testing

- **473 tests green** (467 `ZcapLd.Core` + 6 `ZcapLd.AspNetCore`); full solution builds.
- All Phase 0 goldens remain **byte-identical** across every stage — the guarantee that 3.x-issued capabilities still verify.

## Related

- Tracks zcap-dotnet#108 (evaluation, plan, and decisions).
- Upstream enablers: dataproofs-dotnet#4 (verify-dispatch for legacy proofs) and dataproofs-dotnet#7 (the `DataProofsDotnet.Legacy` suites).


moisesja (commented):
1 — ToDataIntegrityProof round-trip is load-bearing and not fully covered (medium)

LegacyProofCrypto.ToDataIntegrityProof maps zcap's Proof → DataIntegrityProof via a JSON round-trip between two different JsonSerializerOptions. The assumption is that every zcap-specific field (capabilityChain, the embedded delegated capability object, capabilityAction, invocationTarget) survives into AdditionalProperties on DataIntegrityProof and is included in the payload DataProofs builds. The existing golden-vector tests use a simple first-level delegation — the delegated-invocation path with a full parent capability embedded in the proof is not exercised. If DataProofs changes its AdditionalProperties handling or treats a field as "known" (dropping it from AdditionalProperties), zcap-specific proof fields disappear silently from the signing payload. Recommend adding a test that deserializes the DataIntegrityProof.AdditionalProperties for a proof with a multi-entry capabilityChain, capabilityAction, and a nested capability object, and asserts all three appear faithfully.

2 — BuildDocumentElement @context injection (low-medium)

The !TryGetProperty("@context") guard injects the zcap context only if absent. If an Invocation already carries a different @context, the injection is skipped and RDFC expansion silently produces wrong N-Quads. In practice zcap controls the model so this is unlikely, but the code doesn't assert the invariant. Recommend a comment ("Invocation never sets @context in the model; this handles the serialized-without-context case") or a throw if a non-zcap @context is found.

3 — CanonicalizationGoldenVectorTests doc tense (minor)

The class doc describes Stage B as a future swap, but Stage B is complete in this PR. Update to: "Stage B is complete (this PR). These tests now lock the delegating implementations' behavior against future changes."

4 — canonicalizationMethod not validated (low)

SigningService and VerificationService accept an arbitrary string. "rdfc-1.0" (wrong case) silently falls through to JCS in LegacyProofCrypto.Resolve. The consumer gets JCS instead of RDFC-1.0 with no error. Recommend throwing ArgumentException for any value other than "JCS" or "RDFC-1.0" in the constructors or in Resolve.

PR #111, opened 2026-06-16
fix(security): remediate confirmed findings from the security + compliance scan

moisesja:
## Summary

A multi-agent **security + W3C ZCAP-LD compliance scan** produced 28 confirmed findings (5 refuted) after adversarial per-finding verification. This PR fixes every confirmed finding that required a code change. Each finding's severity below is the verified (skeptic-corrected) rating.

## Fixes

**🔴 High**
- **ValidWhileTrue SSRF** — the caveat URI is attacker-controlled. New `SsrfGuard` (pre-flight host check + rebind-safe `ConnectCallback`); `HttpValidWhileTrueHandler` pre-flights every URI; `AddZcapValidWhileTrueSupport` hardens the named `HttpClient` (no auto-redirect, connect-time guard, 10 s timeout, 64 KB response cap). Blocks cloud metadata `169.254.169.254`, loopback, RFC1918/CGNAT/ULA, IPv4-mapped bypass.
- **Delegated `expires` MUST on the verify path** — `VerifyBuiltChainAsync` now rejects a delegated link with no `expires` (`MalformedCapability`), gated `requireDelegatedExpires` (default on; `false` on the revocation-auth path so non-expiring delegations stay **revocable**). Closes the attenuation gap where a child could outlive a short-lived parent by omitting `expires`.

**🟠 Medium**
- `UsageCountCaveat` **fails closed** unless the count is supplied via `InvocationContext.Properties["zcap:usageCount"]` (`CurrentUses` was always 0 on the wire → silently unlimited).
- Offsetless timestamps parse as **UTC** (`AssumeUniversal`); `ParseOrNull` is total.
- `MaxChainLength` enforced **during** the chain walk (bounds work, still surfaces `ChainTooLong`).
- `RdfcDocumentCanonicalizer` reuses `ZcapJsonOptions.Default` so caveats survive into the N-Quads.

**🟡 Low / hardening**
- Standalone proof path enforces parent/child caveat compatibility (symmetric with the chain path).
- `IsValidInvocationTarget` rejects `..` path-segment suffixes.
- Revocation **GET** discloses only `{capabilityId, isRevoked}` to anonymous callers; revoke **POST** caps the body at 1 MB before deserialization.
- `ZcapJsonOptions.Default` sets `AllowDuplicateProperties = false`.
- `SigningService` self-verifies the produced proof against the resolved key (catches HSM/KMS key mix-ups at issuance).
- Built-in time caveats evaluate against `InvocationContext.InvocationTime` (signed `created`).
- Dead `JsonCanonicalizer.RemoveProofField`/`CanonicalizeWithoutProof` removed.

## Deliberately not changed
- **Per-instance nonce/revocation stores** (convenience ctors): inherent to in-memory dev stores; already documented. Production multi-instance must supply shared stores.
- The 5 refuted findings (JCS null-strip differential, `ConcatenateHashes` naming, the "unbounded chain DoS" framed as High, case-variant keys, malformed-`expires` classification) were disproven by the skeptics.

## Test plan
- `dotnet build ZcapLd.sln` — succeeds.
- `dotnet test ZcapLd.sln` — **520 passing, 0 failing** (487 Core + 33 ASP.NET), incl. **41 new regression tests** (`SsrfGuardTests`, `ZcapTimestampsTests`, dot-segment cases, `UsageCountCaveat_NoContextCount_FailsClosed`, `SignedRevoke_DelegatedWithoutExpires_*`).
- `dotnet run --project examples/ZcapLd.Examples` — full demo runs clean.
- Existing tests updated to assert the new (correct) behavior rather than being silenced.

## Behavior-contract changes for consumers
- A delegated capability **MUST** carry `expires` to verify (it stays revocable).
- `UsageCountCaveat` requires the count via `InvocationContext.Properties["zcap:usageCount"]` and fails closed otherwise.
- The anonymous revocation status endpoint no longer returns `RevokedBy`/`Reason`/`Metadata`/`RootCapabilityId`.


moisesja (commented):
Status: Looks good, 3 breaking changes need release note callout

CI: Both build-test-pack-core and build-test-pack-aspnet — green. 520 tests passing, 41 new.

Strengths:

SsrfGuard two-layer defense (pre-flight DNS + connect-time ConnectCallback) is the right architecture; IPv4-mapped bypass caught correctly
UsageCountCaveat fail-closed fix addresses a real enforcement gap (the wire-deserialized CurrentUses was always 0)
requireDelegatedExpires: false carve-out on the revocation path follows the established #73/#99 pattern consistently
ConfirmSignature self-verify costs only a local CPU operation (not an HSM round-trip)
Path traversal check correctly distinguishes .. as a complete segment from a..b and ?p=../x
Breaking changes that need release notes:

UsageCountCaveat — callers who set CurrentUses directly will now get denied; must migrate to InvocationContext.Properties["zcap:usageCount"]
Delegated expires MUST — any in-the-wild delegated zcap missing expires will fail with MalformedCapability; re-delegation required before upgrading
Revocation GET response — RevokedBy/Reason/RootCapabilityId fields are gone from the anonymous endpoint; consumers need their own authorized endpoint for the richer payload
Minor observations: Double DNS lookup in ValidWhileTrue (pre-flight + connect-time) is absorbed by the resolver cache in practice; PooledConnectionLifetime = 2 min is fine but worth confirming for expected traffic volume.

moisesja:
Addressed the review in `35b512e`.

**Breaking-change release notes** — added to `CHANGELOG.md` under the unreleased section (a new "From the security + compliance scan remediation" sub-list in **Breaking Changes**, plus a **Security** section). All three callouts included with migration guidance:
- Delegated `expires` MUST → re-delegation required for in-the-wild zcaps missing it (revoke path exempt).
- `UsageCountCaveat` → supply the count via `InvocationContext.Properties["zcap:usageCount"]`; direct `CurrentUses` callers are now denied.
- Revocation GET → returns only `{capabilityId, isRevoked}`; richer payload needs a consumer-owned authorized endpoint.

**Minor observations** (no code change, per your read):
- *Double DNS lookup* — intentional: the pre-flight is a fail-fast that also protects consumers who wire their own `HttpClient` without the connect-time callback; the second resolve is the authoritative rebind-safe one and is absorbed by the resolver cache, as you noted.
- *`PooledConnectionLifetime = 2 min`* — kept; it only bounds pooled-connection reuse for these low-volume status probes and is independent of correctness. Happy to make it (and the timeout / size caps) configurable on `AddZcapValidWhileTrueSupport` in a follow-up if you'd prefer.

